PR #6, opened 2015-12-28
Ignore the "Class already exists" error in multiple-window application

kondrak:
Since I need a second window now, here's the (possibly easiest) way to fix the duplicate class problem.

### DIFF
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -249,8 +249,14 @@ impl Window {
             };
 
             if user32::RegisterClassW(&class) == 0 {
-                println!("Unable to register class, error {}", kernel32::GetLastError() as u32);
-                return None;
+                let err_code = kernel32::GetLastError() as u32;
+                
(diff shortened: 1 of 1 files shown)


src/windows.rs
```diff
@@ -249,8 +249,14 @@ impl Window {
             };
 
             if user32::RegisterClassW(&class) == 0 {
-                println!("Unable to register class, error {}", kernel32::GetLastError() as u32);
-                return None;
+                let err_code = kernel32::GetLastError() as u32;
```
emoon: Can just do

```Rust
if kernel32::GetLastError() as u32 != 1410 {
  ...
}
```

Also keep brace on the first line (I try to have the code following the Rust guidelines for code https://github.com/rust-lang/rust/blob/master/src/doc/style/style/braces.md )

kondrak:
Re-formatted and commited.

emoon:
Thanks!

PR #14, opened 2016-01-19
Include needed modules on the example code

bltavares:
The example code on `README.md` requires the `use` line to have the modules available when compiling.
Without it the example can't compile.


emoon:
Good catch! Thanks.

bltavares:
You are welcome (:

PR #20, opened 2016-06-14
Create window.update_title

jedahan:
This function lets you update the window title.

I only got it working with OSX, since I have nothing else to test on.


emoon:
Thanks! But as is I can't accept this PR as is. It would be great if you just added a dummy implementation here

https://github.com/emoon/rust_minifb/blob/master/src/os/unix/mod.rs#L295

and here

https://github.com/emoon/rust_minifb/blob/master/src/os/windows/mod.rs#L480

Which would make it build for the other platforms (and code would be callable but not doing anything)

Then I can accept the PR and do the implementation of it.

jedahan:
Cool, i'll get on that tonight after work :)

emoon:
I will close this PR now because I'm implementing this myself over here https://github.com/emoon/rust_minifb/tree/cursor-title-support (notice that I also decided to call the function ``set_title`` instead of ``update_title``) I'm also adding support to change the cursor to a number of predefined ones that are support on Mac, Linux and Windows. Hopefully a new version (0.8.0) will be out today.

emoon:
Version 0.8.0 has now been released

jedahan:
Thanks so much!

PR #37, opened 2017-07-08
Support period key on *nix.

kondrak:
Found this while I was looking at the backspace issue on Linux - period key is not handled and it's something that my emulator uses, too!


emoon:
Thanks!

PR #41, opened 2017-08-15
Fix Redox build

xTibor:
A typo in the new bounds checking code (900deba9a7851417c0253b57feba4a089f18b12a) broke the Redox port.


emoon:
Doh! I tried to be careful but I guess I typed wrong here :) Thanks for the fix. I will release a new version with it later today.

emoon:
0.10.1 has now been released

xTibor:
Thanks!

PR #43, opened 2017-10-04
Fix Travis-CI build

AlexEne:
This change should fix the failing builds on travis-ci.
I briefly tested it on my account and everything seems to work.

Not sure if this is the proper way to fix it since anyone who consumes this library and builds on travis ci needs to set the script to apt-get that pkg anyway, but I suppose it's better to see it in the config for the main repro. 

Anyway it's way less painful than rust SDL on windows for example.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,5 @@ rust:
 matrix:
   allow_failures:
     - rust: nightly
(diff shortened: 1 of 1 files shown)


emoon:
Thanks!

PR #49, opened 2018-02-15
Cargo.toml: Remove homepage link

jonas-schievink:
The `homepage` key is meant for dedicated library websites, a link to the repository is already provided by the `repository` key.


emoon:
Makes sense. Thanks!

PR #56, opened 2018-08-09
Add debug impls

LukasKalbertodt:
Hi there!

I came here to add missing `Debug` impls. While doing so I noticed a few strange things which I tried to fix in the second commit. I don't quite understand why some modules were public but `doc(hidden)` (e.g. `pub mod os;`) and why some public items were `doc(hidden)` despite being used (e.g. `Key` and `Result`).

I hope these changes are OK! If not, just tell me what to change (I can also just remove some of the commits).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -87,18 +87,13 @@ pub trait InputCallback {
     fn add_char(&mut self, uni_char: u32);
 }
 
-#[doc(hidden)]
 mod error;
 pub use self::error::Error;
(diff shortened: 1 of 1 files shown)


emoon:
Good stuff! Thanks :) I will publish a new version with these fixes soon.

emoon:
0.10.7 has now been published.

LukasKalbertodt:
Awesome, thank you :)

PR #62, opened 2018-12-18
Added Window.is_key_released

tdaffin:
I'm looking into making a Rust version of @Javidx9 's olcPixelGameEngine: https://github.com/OneLoneCoder/olcPixelGameEngine
Not much progress yet, but a good way to learn Rust for sure!

Anyhow, so far it seems that minifb is the nearest thing I can find that does the low level portion of what I'm trying to do, except for the fact that it doesn't let you determine if a key has just been released.

This PR adds the is_key_released method to address that shortfall.

I guess that the version could be bumped to 0.12.0 as I'm adding a new function, but I figured you could do that if you felt it was in keeping with what you're doing.

### DIFF
--- a/src/key_handler.rs
+++ b/src/key_handler.rs
@@ -8,6 +8,7 @@ pub struct KeyHandler {
     prev_time: f64,
     delta_time: f32,
     keys: [bool; 512],
+    keys_prev: [bool; 512],
     keys_down_duration: [f32; 512],
     key_repeat_delay: f32,
     key_repeat_rate: f32,
@@ -18,6 +19,7 @@ impl KeyHandler {
         KeyHandler {
             key_callback: None,
             keys: [false; 512],
+            keys_prev: [false; 512],
             keys_down_duration: [-1.0; 512],
             prev_time: time::precise_time_s(),
             delta_time: 0.0,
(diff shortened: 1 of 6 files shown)


tdaffin:
Not sure why the appveyor build failed -- something to do with the 'cc' dependency?

emoon:
Hey, Thanks for the PR. it looks good! Not sure what is going on with Appveyor 🤔 I will try to investigate it and get back to you.

emoon:
The issue with `cc` being broken is that Appveyor is using a very old (1.16) version of the the Rust compiler so this isn't really an issue in practice but somewhat annoying

tdaffin:
Ok thanks!
Looks like Appveyor is using the correct beta and nightly, which are now succeeding, but for some reason their stable is not picking up the latest.

I took a quick look at the appveyor.yml and appveyor_rust_install.ps1 and they ultimately depend on the file returned by https://static.rust-lang.org/dist/channel-rust-stable which is returning rust-1.16.0
I wonder if it would be possible to change the url defined in the powershell script to point to the real stable builds?

tdaffin:
Perhaps updating the appveyor.yml would fix things.  I found an example [here](https://github.com/starkat99/appveyor-rust/blob/master/appveyor.yml) that uses rustup to install the right version of rust.

emoon:
Yeah I guess that would be possible but it would be better if that url would point to the correct info :) otherwise one would have to change the script when the stable release is updated.

emoon:
looks much better yes

PR #64, opened 2018-12-19
Bump version to 0.11.2

tdaffin:
I'd like to make the recent 'is_key_released' feature available in the crate on crates.io, and a version bump  seems necessary.
Is this also sufficient, or is there something else that has to be done to get the new version on crates.io?


tdaffin:
Thanks!

PR #65, opened 2018-12-28
Add additional derived traits to Key enum

stegmanh:
This PR simply adds the `Eq, Ord, PartialOrd and Hash` traits to the derive on top of the Key enum. It allows Keys to be stored in Sets, among other QOL things.


emoon:
Looks good. Thanks!

Issue #66, opened 2019-01-13
Resize handling on X11 causes buffer overruns, segfaults.


cbiffle:
This can be reproduced easily by running the `noise` or `title_cursor` demos under a tiling window manager. Both these demos set `resize: true` but draw to a 32-bit buffer, which is a combination that doesn't look officially supported. Tiling window managers tend to resize windows immediately when they're opened. (I bet you can repro this under the default Ubuntu window manager by resizing the window manually.)

However, setting `resize: false` doesn't actually fix it, because the flags aren't stored anywhere in the window info. They're only set on the X11 hints, which are just that -- hints. The window manager can still resize the client's window.

And when it does, the following happens:

- The `ConfigureNotify` event gets processed [here](https://github.com/emoon/rust_minifb/blob/master/src/native/x11/X11MiniFB.c#L419), changing the `width` and `height` without resizing the `image` or `display_buffer`.
- At the next call to `mfb_update_with_buffer` the new size is passed *along with the old buffer* to a copying routine, e.g. [this one](https://github.com/emoon/rust_minifb/blob/master/src/native/x11/X11MiniFB.c#L706).
- The copying routine happily writes the buffer as though it were the new size, regardless of its actual size. (With the allocator on my machine, this happens to stomp the WindowInfo struct itself, making debugging quite confusing.)

I can see two ways of fixing this.

The "minimal" way would be to record the fact that the window doesn't support resizing somewhere, and (if that flag is set) never change width and height after they are initially set. This means that resizing the window just adjusts a black rectangle around the original image, but at least it doesn't segfault. I've got a slightly hacked patch for this if you want.

The "thorough" way would be to reallocate the buffers on `ConfigureNotify` events. This would enable basic resize support when drawing to a 32-bit buffer. I bet there's more to this, though -- I'm not familiar with your library, particularly on non-X11 platforms.

emoon:
Thanks for the report. I will take a look at what makes sense doing.

emoon:
Sorry I haven't been able to look into this yet :(

cbiffle:
This is a potentially exploitable safety issue, consider yanking the package or posting a warning for X11 users.

emoon:
Maybe I can take your slightly hacked patch to begin with if that is fine?

PR #68, opened 2019-03-14
raise shared_data update

FauxFaux:
This helps the `check_buffer_size` run against the actual window size, which stops us from segfaulting if the window size has changed unexpectedly, e.g. because a window manager is evil.

This improves #66, but by no means fixes it (the examples still *panic* on start here (i3), they just don't segfault).


emoon:
Thanks! @cbiffle can you test if this also doesn't segfault for you?

PR #69, opened 2019-03-20
Update winapi

### LINKED ISSUE #53: Move to winapi 0.3
The 0.3 version lets you bind and compile against just the modules that are actually used, cuts down on compile times quite a bit.

zxey:
Updates winapi to version 0.3.
Removes user32-sys, kernel32-sys and gdi32-sys, they are no longer needed, because they are provided by winapi-rs.

Also removed star glob imports, some of the imported symbols are quite generic and this way it is also more consistent with the rest of the code.

Closes #53

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,10 +22,10 @@ cc = "1.0"
 time = "0.1.34"
 
 [target.'cfg(windows)'.dependencies]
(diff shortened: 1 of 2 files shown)


emoon:
Looks good! Thanks.

PR #70, opened 2019-03-29
Rewrite linux backend in rust

### LINKED ISSUE #11: Rewrite Linux/X11 version in Rust

FauxFaux:
Fixes #11.

I rebased @andwj's port on top of master, and fixed some bugs. It seems to work for me. It does not address any issues with tiling/reparenting window managers resizing non-resizable windows, but this just causes the examples to panic ("Update failed because input buffer is too small."), no segfault, as before.

#11 talks about multiple windows not working, but I can't reproduce that: it works for me, in the trivial example.
I understand why you would only want to open the display once per process, but I can't immediately see anywhere that says you shouldn't...
Sharing `DisplayInfo` via. `lazy_static` is problematic as it must be `Sync`, and declaring some of those fields `Sync` is .. probably not a good idea. I'd rather see the library move towards making the user create a `Minifb`, then using that to create windows?

I merged the scaling code, but haven't really tested it. 2x works!


emoon:
Awesome! Thanks a lot :)

Issue #71, opened 2019-03-31
mouse problematic with odd-sized windows and scale


FauxFaux:
If your window `width` is not evenly divisible by `scale` (i.e. 463 pixels for Scale2x), something odd is going on with the `mouse` example, on unix. I haven't tried other platforms.

The resolved pixel appears "further along" the buffer than expected, as if multiplying by the (scaled) width is incorrect.

Is this a bug in the example, or is it a bug in the buffer handling in unix?

This is easier to test after (or caused by!) #72.

PR #72, opened 2019-03-31
x11 and examples: resize

### LINKED ISSUE #66: Resize handling on X11 causes buffer overruns, segfaults.

FauxFaux:
Handle, to some extent, resizing under x11.

The middle of the event loop is perhaps not the best place to reallocate the image, but it is the simplest. Specifically, there are issues with error handling; at the moment, it just panics (`expect`).

I also fiddled the examples, `noise` and `mouse` now support resizing properly (i.e. they handle updating the buffer), and `multi` and `title_cursor` ask for non-resizable windows.

This should hopefully finally fix #66. It may introduce #71.

### DIFF
--- a/examples/mouse.rs
+++ b/examples/mouse.rs
@@ -10,6 +10,7 @@ fn main() {
 
     let mut window = match Window::new("Mouse Draw - Press ESC to exit", WIDTH, HEIGHT,
                                        WindowOptions {
+                                           resize: true,
                                            scale: Scale::X2,
                                            ..WindowOptions::default()
                                        }) {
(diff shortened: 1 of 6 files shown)


emoon:
Thanks!

PR #79, opened 2019-08-22
unix: replaced scale functions with macro and added missing invocations

johannst:
Hi,

just run into following limitation when using minifb, hope this enhancement can help others.

Best
Johannes


emoon:
Thanks for the fix!

PR #89, opened 2019-09-26
Improve documentation for method `update_with_buffer` of `Window`

### LINKED ISSUE #88: Specify encoding of RGBA colorspace in documentation
The `update_with_buffer` method of `Window` takes a `&[u32]`. That's a byte each for red, green. blue, and alpha. But the specific encoding of the pixel isn't specified in the [docs](https://docs.rs/minifb/0.13.0/minifb/struct.Window.html#method.update_with_buffer). It took a little trial and error to find that the correct encoding for me on Unix is ARGB:

```rust
fn from_u8_rgba(r: u8, g: u8, b: u8, a: u8) -> u32 {
    let (r, g, b, a) = (r as u32, g as u32, b as u32, a as u32);
    b | (g << 8) | (r << 16) | (a << 24)
}
```
```rust
// Color the window magenta
let buffer: Vec<u32> = vec![from_u8_rgba(255, 0, 255, 255); 640 * 400];
let mut window = Window::new("Test", 640, 400, WindowOptions::default()).unwrap();
window.update_with_buffer(&buffer).unwrap();
```

I could make a pull request to add to the docs that the encoding is ARGB but I'm not sure if this is platform specific or not.

ArifRoktim:
Closes #88 

I also added just to the example, a function (`from_u8_rgb`) that shows how to properly encode the RGB value. If you want, this function could be moved to the crate itself.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -222,14 +222,23 @@ impl Window {
     }
 
     ///
-    /// Updates the window with a 32-bit pixel buffer. Notice that the buffer needs to be at least
-    /// the size of the created window
+    /// Updates the window with a 32-bit pixel buffer. The encoding for each pixel is `0RGB`:
+    /// The upper 8-bits are ignored, the next 8-bits are for the red channel, the next 8-bits
+    /// afterwards for the green channel, and the lower 8-bits for the blue channel.
+    ///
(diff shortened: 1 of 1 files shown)


emoon:
Thanks! I will take a look at this later today

emoon:
Thanks!

Issue #47, opened 2018-01-19
WASM support


tversteeg:
Would it fit the project to add a WebAssembly target? 

- Creating a simple HTML5 canvas object and then copying the buffer of pixels every frame to JavaScript and sending the events back. Or it could be done using [`ImageData`](https://www.hellorust.com/demos/feistel/index.html).
- Scaling could be implemented like [this](https://jsfiddle.net/namuol/VAXrL/1459/).
- A example for keyboard input is shown [here](https://aochagavia.github.io/blog/rocket---a-rust-game-running-on-wasm/).
- The menus could be a problem but that could be fixed with custom buttons in JavaScript.

I would love to help with developing/testing this if it fits your vision of the project!

emoon:
That is an interesting question. I haven't actually thought about that. I have very little experience with web and esp WebAssembly but if you would want to implement this version I would be very happy with including it in the master branch.

As for menus Linux doesn't also support menus because there is no native API for doing them so I think that is fine.

FloVanGH:
I've created a window wrapper for wasm based on stdweb https://github.com/koute/stdweb. I use it for OrbTk https://gitlab.redox-os.org/redox-os/orbtk. Maybe I could help to implement wasm support.

tversteeg:
I've made a small attempt to set this up but I got stuck quite quickly. It seems you are well versed in WASM, maybe we could implement this feature together?

FloVanGH:
@tversteeg sure why not. Do you use a specific library? I uses stdweb (https://github.com/koute/stdweb), it's very comfortable for interacting with web api on Rust wasm. You could find my wasm window shell wrapper on https://gitlab.redox-os.org/redox-os/orbtk/blob/master/crates/shell/src/web/mod.rs. The code could need a little bit refactoring and there are also OrbTk specific parts, but we could use it as example.

I also create a tool called cargo-node https://github.com/FloVanGH/cargo-node to build, deploy and run rust wasm applications as electron or cordova app. It could help us to test the port and we could use it also for a initial android port until there is a native port.

At the moment I have a lot of work todo on OrbTk but I can spent some time in a few weeks. I hope after the port is finished I can replace my wasm window shell wrapper.

tversteeg:
I've created a small start here: #92 
It's compiling but not working at all at the moment, I'm not sure what I'm doing wrong.

FloVanGH:
[...]

PR #92, opened 2019-10-12
#47 WIP wasm support

tversteeg:
Work in progress for WebAssembly support. It compiles at the moment but it's non-functional.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ cc = "1.0"
 [dependencies]
 cast = "0.2"
 time = "0.1.34"
+instant = { version = "0.1", features = [ "stdweb", "now" ] }
 
 [target.'cfg(windows)'.dependencies.winapi]
 version = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,3 +52,12 @@ x11-dl = "2.18.3"
 
 [target.x86_64-unknown-redox.dependencies]
 orbclient = "0.3.20"
+
+[target.wasm32-unknown-unknown.dependencies]
+stdweb = "0.4.20"
+
+[target.wasm32-unknown-emscripten.dependencies]
+stdweb = "0.4.20"
+
+[target.asmjs-unknown-emscripten.dependencies]
+stdweb = "0.4.20"
--- a/Web.toml
+++ b/Web.toml
@@ -1,0 +1,8 @@
+# The default value of `--target` used when building this crate
+# in cases where it's not specified on the command line.
+default-target = "wasm32-unknown-unknown"
+
+[cargo-web]
+# Asserts the minimum required version of `cargo-web` necessary
+# to compile this crate; supported since 0.6.0.
+minimum-version = "0.6.0"
--- a/src/key_handler.rs
+++ b/src/key_handler.rs
@@ -1,7 +1,8 @@
+extern crate instant;
 extern crate time;
 
 use std::mem;
-use {Key, KeyRepeat, InputCallback};
+use {InputCallback, Key, KeyRepeat};
 
 pub struct KeyHandler {
     pub key_callback: Option<Box<dyn InputCallback>>,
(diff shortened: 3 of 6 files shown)


tversteeg:
@emoon the WASM VM is throwing a panic on `time::precise_time_s` (in `KeyHandler::new`). Apparently this is not supported in WASM. Would it be fine if I would replace that code with the [instant library](https://github.com/sebcrozet/instant)?

emoon:
@tversteeg sure as long as it provides the same functionality (and only has few/no dependencies) it's fine

tversteeg:
I'm kind of stuck with this PR.

The problem is that HTML5 has its own event loop for drawing on the canvas using [`window.requestAnimationFrame(callback)`](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame). I can't figure out a way to tie this event loop with the user defined `while` loop rust_minifb uses.
Normally I would maintain 2 threads and synchronize the buffer between them but that's not possible since WASM doesn't support threading yet.

I kind of need some help with this problem.
@FloVanGH?
@sinclairzx81?

FloVanGH:
@tversteeg I had the same problem on OrbTk. Therefore I've created a Runner struct where you can define a custom loop for wasm. It is not the best code but at the moment it works. Maybe for wasm support you would provide a similar concept.

Wasm Runner:
https://gitlab.redox-os.org/redox-os/orbtk/blob/master/crates/shell/src/web/mod.rs#L224

Default Runner:
https://gitlab.redox-os.org/redox-os/orbtk/blob/master/crates/shell/src/web/mod.rs#L224

Winit solved that problem by using an event loop:

https://github.com/rust-windowing/winit/blob/master/examples/window.rs

sinclairzx81:
@tversteeg Hi,

Thanks for the mention. To be honest, I am not too clued up on the best approach to take here that would align with the current `while(window.is_open()) {}` loop given current browser capabilities, but I think @FloVanGH ' s suggestion seems like a pretty good one. Having the JavaScript host call into the WASM context on `requestAnimationFrame(...)` with the `window.request_animation_frame(move |window| {... window.update_with_buffer(..); })` closure invoked on each call seems like it would work quite well.

Just on the threading side; it might be possible to run minifb within a web worker at some point in the  future. Sadly I don't think this is quite ready for showtime as `SharedArrayBuffer` has been currently disabled in most browsers in response to [Spectre](https://meltdownattack.com/) and the only way i know would be to copy buffers across contexts (super slow). However, this functionality should land one day, and should (in theory at least) allow one to retain the `while` loop simplicity through the use of JS Atomics.

I've included some MDN links for `SharedArrayBuffer` and `Atomics` below for reference. Due to much of this functionality currently being disabled in most browsers, I believe it's still possible to experiment with the threading via NodeJS and its [worker_thread](https://nodejs.org/api/worker_threads.html) module which does allow for `SharedArrayBuffer` and `Atomics`, with perhaps [Atomics.wait](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/wait) and [Atomics.notify](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/notify) being of some use to help synchronize without explicitly needing to run as event driven closures. Will take some poking around to get it all to work, but would allow the WASM implementations of minifb to align with the Desktop implementations (and keep the 'write one, run everywhere' nature of minifb in check)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer

Also, great work on the WASM support!

tversteeg:
Thanks a lot both! That's really helpful. I don't think I'll have much time this weekend for it but I'll definitely take a good look next week.

tversteeg:
@FloVanGH 
I've been taking a look at the code you supplied (sorry for being late, I was very busy), and that would entail that the whole `while` loop of minifb needs to be changed and put inside a closure right? I doubt that @emoon would want this because keeping the `while` loop seems to be one of the strong points of minifb.

tversteeg:
I'm closing this PR. Unfortunately I can't find a solution where the current structure is maintained. If anyone has any hints or the SharedArrayBuffer from @sinclairzx81's comment is widely available I'll re-open it.

emoon:
I see. That is too bad. Like I said if we could have a slightly different API when using WASM I'm open for that idea as well.

FloVanGH:
@tversteeg I think I could help. But at the moment I have a lot to do on OrbTk. On my next free time slot I will try to help you with this or.

PR #96, opened 2019-10-28
fix 'attempt to subtract with overflow' bug

citorva:
A program which creates a window with this library when the requested window size is taller than the screen size will crash because the two following lines do not take into this case.
```rust
let x = (d.screen_width - width) / 2;
let y = (d.screen_height - height) / 2;
```

The patchwork assigns this processed value when the window size is smaller than the screen size or 0 else.


emoon:
Thanks! I'm going to go over all of this in more detail during a break I have coming up to try to make things a bit more reliable.

PR #98, opened 2019-11-02
Fix window close handling

nbdd0121:
Currently we interned `WM_DELETE_WINDOW` but does not call `XSetWMProtocols` to listen to it, causing a fatal error when closing the window by clicking "X". This patch fixes the issue.


emoon:
Thanks!

PR #128, opened 2019-12-31
Restructure Unix support(Wayland prep)

nifker:
I didnt use git mv that why the changes are so "big"


emoon:
Hum. I'm a bit unsure about this change. I would prefer to have the code separate so there is no need to do `switch *self {` for each call to minifb. My suggestion would be to under `https://github.com/emoon/rust_minifb/tree/master/src/os/unix` have an `x11.rs` a `wayland.rs` and then the code that is shared between the two split out into some `unix common` code instead.

nifker:
The only code which will be in common will be "mouse_x, mouse_y, buttons, etc". Do you mean that? But there will still be many methods/function which cant be shared.

emoon:
ok, if very little can be shared I think it's better to have them as fully separate backends and if you find some code that is "large:ish" that can be shared just put that code in a separate file and call into that from both backends (`x11` and `wayland`) a little bit of duplication isn't a big deal tho.

nifker:
Thats the current struct I could have come up with, but I think draw_buffer would  not be needed on Wayland.

```
//Common window attributes between X11 and Wayland
struct Window{
	draw_buffer: Vec<u32>,
	
	width: u32,
	height: u32,
	
	scale: i32,
	bg_color: u32,
	scale_mode: ScaleMode,
	
	mouse_x: f32,
	mouse_y: f32,
	scroll_x: f32,
	scroll_y: f32,
	buttons: [u8; 3],
	prev_cursor: CursorStyle,
	
	should_close: bool,
	
	key_handler: KeyHandler,
	update_rate: UpdateRate,
	menus: Vec<UnixMenu>,
	menu_counter: MenuHandle,

	specific: SpecificWindow
}


//Differentiate between Wayland and X11 at run-time
pub enum SpecificWindow{
	X11(x11::Window),
	Wayland(())//WlWindow
}
```

emoon:
I would flip this around instead. Something like

```Rust
struct CommonUnixData {
width: u32,
	height: u32,
	
	scale: i32,
	bg_color: u32,
	scale_mode: ScaleMode,
	
	mouse_x: f32,
	mouse_y: f32,
	scroll_x: f32,
	scroll_y: f32,
	buttons: [u8; 3],
	prev_cursor: CursorStyle,
	
	should_close: bool,
	
	key_handler: KeyHandler,
	update_rate: UpdateRate,
	menus: Vec<UnixMenu>,
	menu_counter: MenuHandle,
}

struct X11Window {
   common_unix: CommonUnixData;
   // x11 data goes here
}

struct WaylandWindow {
  common_unix: CommonUnixData;
   // Wayland data goes here
}
```

Then in lib.rs

```Rust
#[cfg(any(
    target_os = "linux",
    target_os = "freebsd",
    target_os = "dragonfly",
    target_os = "netbsd",
    target_os = "openbsd"
))]
use self::os::unix as imp;
#[cfg(target_os = "windows")]
use self::os::windows as imp;
///
/// Window is used to open up a window. It's possible to optionally display a 32-bit buffer when
/// the widow is set as non-resizable.
///
pub struct Window(imp::Window);
```

The above would need to be change so when Wayland is used so `use self::os::unix as imp;` uses the `self::os::unix::wayland` and x11 is `self::os::unix::x11` or something like that

nifker:
Oh yes, that makes more sense.

nifker:
Wait the thing is when we put the common data into waylands/x11s own struct which struct is the new Window struct then?

nifker:
> The above would need to be change so when Wayland is used so `use self::os::unix as imp;` uses the `self::os::unix::wayland` and x11 is `self::os::unix::x11` or something like that
We cant evaluate if the user runs X11 or Wayland at compile time.

emoon:
Oh. I thought we would know at compile time. Yeah then it makes things a bit harder. Then we would need to do a trait or have a run-time check (as you suggested first) in the trait idea we would then have

```Rust

train UnixWindow {
   // functions such as get_mouse, etc, here
}

impl UnixWindow for X11Window {
  ...
}

impl UnixWindow for WaylandWindow {
   ...
}

struct Window {
  imp: Box<WindowTrait>, // Not sure if we need to box it?
}
```

Then when creating the Window we check which implementation to use and instantiate that one.

nifker:
No we cant know this because a user might switch between X11 environments and Wayland or even run a Wayland compositor inside a X11 environment.
I mean we could evaluate this at compile time but it makes no sense(because it would be dependent on the developers environment).
Yes this for sure needs to be boxed(I dont think an Arc makes sense here).

nifker:
I did some basic with enums and traits and the enum version was always faster. So I think it would be better to implement this with an enum.
The box also introduces the "issue" that we now have the memory on heap and the X11 struct stores data on the heap too which could slow down.

emoon:
Alright lets go for enum version then. I would suggest at least to try to keep the x11 code and wayland code in two separate files. And just have the `unix/mod.rs` call into each implementation. This will make it a bit cleaner. 

Also having data on the heap or stack doesn't matter in this case in terms of performance as it's just one alloc at the start (when creating the window)

nifker:
But we have allocations when the window size increases or not?
Are the UnixMenus specific to X11 or is it some standard?

nifker:
Ok I now just need to move some functions which can be properly called from CommonWindowData.

emoon:
I looked at the changes now and I changed my mind. Keep the code fully separated for now. I think the common data thing just makes things the code more compilated

nifker:
You mean like a dropdown-menu?
Ill be then reverting back to "8f104a993ff28071ee9f3cda96ae46cf48ab38a1".

emoon:
Yes, On Windows you have a menu bar on the Window while on macOS you have it at the top of the screen that is tied to the active application.

nifker:
As I have no clue how to properly revert the commits I just changed the files.

emoon:
no problem with the no revert / change. I will just squash all of these changes into one later on

nifker:
Why does the linter actually fail I cant find any concrete error there :O

emoon:
do you have some custom rust fmt somewhere?

nifker:
Not that I know of any global settings I applied to it.

emoon:
it's weird because for you it looks like it puts 8 wide tabs on everything (when it should 4 four spaces)

nifker:
I use hard tabs in my editior, even though Ive ran cargo fmt(as you can see) after that.

emoon:
I pushed a generated default config now. Hopefully that fixes it

nifker:
Ill apply it right now.

nifker:
Seems like I need to downgrade my rustfmt for that, not sure too sure how to do that.

emoon:
Hum.. so nightly rustfmt doesn't read the config? that sounds strange.

nifker:
@emoon https://github.com/rust-lang/rustfmt/issues/3997#issuecomment-571191668

emoon:
I have removed the required version setting now. Hopefully it works now

nifker:
Seems to be successfull now.

nifker:
Is there something missing for the merge?

emoon:
Sorry. I didn't know if you wanted to do some more changes. I will review now and merge if ok

PR #141, opened 2020-02-05
[WIP] Wayland support

nifker:
Implemented:

- [x] new()

- [x] set_title()

- [x] set_background_color() saves only the color - does currently nothing

- [x] is_open()

- [x] get_size()

- [x] get_keys()

- [x] get_keys_pressed()

- [x] get_mouse_pos()

- [x] get_mouse_down()

- [x] get_unscaled_mouse_pos()

- [x] get_scroll_wheel()

- [x] is_key_down()

- [x] set_position()

- [x] set_rate()

- [x] set_key_repeat_rate()

- [x] set_key_repeat_delay()

- [x] is_key_pressed()

- [x] is_key_released()

- [x] update_rate()

- [x] is_active() - keyboard or mouse or both?

- [x] update()

- [x] update_with_buffer_stride() 

- [x] set_cursor_style()

- [x] set_input_callback()

- [x] all menu functions

BUGS:
- keyboard input doesn't work anymore when switching between Wayland and X11 session(Sway specific/wlroots(see: https://github.com/swaywm/sway/issues/3244))


nifker:
It seems like this causes a SIGSEGV: https://smithay.github.io/wayland-rs/src/wayland_cursor/lib.rs.html#114-120 - can someone test this?

cole-h:
I cannot run any example due to 19a590a0425306dea3ef8f67461fa6737d817c1a
(bisected) on sway `1.4-46561fc9 (Feb  7 2020, branch 'master')` and using
`wlroots 0.10.0.r7.g4b051aa9-1`. Reverting those changes fixes the problem. The
`wayland-client` examples don't use `damage_buffer` -- only `attach` and then
`commit`. Potentially unnecessary?

Have you considered using `smithay-client-toolkit` instead of directly using
the wayland-* libraries?

<details><summary>RUST_BACKTRACE=1 from example julia</summary>

```
thread 'main' panicked at 'Cannot send request damage_buffer which requires version >= 4 on proxy wl_surface@5 which is version 1.', /home/cole/.cargo/registry/src/github.com-1ecc6299db9ec823/wayland-client-0.24.1/src/proxy.rs:84:13
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:84
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:61
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1025
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1426
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:65
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:50
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:193
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:210
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:471
  11: rust_begin_unwind
             at src/libstd/panicking.rs:375
  12: std::panicking::begin_panic_fmt
             at src/libstd/panicking.rs:326
  13: wayland_client::proxy::Proxy<I>::send
             at /home/cole/.cargo/registry/src/github.com-1ecc6299db9ec823/wayland-client-0.24.1/src/proxy.rs:84
  14: wayland_client::protocol::wl_surface::WlSurface::damage_buffer
             at ./target/debug/build/wayland-client-d33ae94451f4f1ce/out/wayland_api.rs:5713
  15: minifb::os::unix::wayland::DisplayInfo::new
             at src/os/unix/wayland.rs:125
  16: minifb::os::unix::wayland::Window::new
             at src/os/unix/wayland.rs:312
  17: minifb::os::unix::Window::new
             at src/os/unix/mod.rs:31
  18: minifb::Window::new
             at src/lib.rs:221
  19: julia::main
             at examples/julia.rs:13
  20: std::rt::lang_start::{{closure}}
             at /rustc/5e1a799842ba6ed4a57e91f7ab9435947482f7d8/src/libstd/rt.rs:67
  21: std::rt::lang_start_internal::{{closure}}
             at src/libstd/rt.rs:52
  22: std::panicking::try::do_call
             at src/libstd/panicking.rs:292
  23: __rust_maybe_catch_panic
             at src/libpanic_unwind/lib.rs:78
  24: std::panicking::try
             at src/libstd/panicking.rs:270
  25: std::panic::catch_unwind
             at src/libstd/panic.rs:394
  26: std::rt::lang_start_internal
             at src/libstd/rt.rs:51
  27: std::rt::lang_start
             at /rustc/5e1a799842ba6ed4a57e91f7ab9435947482f7d8/src/libstd/rt.rs:67
  28: main
  29: __libc_start_main
  30: _start
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

</details>

I can also confirm SIGSEGV when exiting any example.

Things I noticed:
- The `title_cursor` example shows a "corrupted" cursor on the right half of
  the window
- `ESC` to exit does not work on examples with that instruction in their window
  title
- The `julia` example does not update until the window is resized -- constantly
  resizing the window shows the animation works, as well as continues animating
  even if it doesn't show (e.g. it would "skip" frames if I were to resize,
  stop, then resize again).

While not important right now, it would greatly improve readability if you were
to run `cargo fmt`/`rustfmt` on your code.

---

Sorry it took so long to try this out and get back to you. Thanks for
spearheading the Wayland implementation!

nifker:
I found wayland-rs quite easy to work with, I only found out about "smithay-client-toolkit" later, I take it into consideration to use it.
Ill take a look at these examples and try to fix them.

nifker:
Ok reversing the issue didnt solve the issue it made it worse, but I forgot the proper version for wl_compositor.

nifker:
@emoon From where do these functions come from(library)?
https://github.com/emoon/rust_minifb/blob/master/src/os/unix/x11.rs#L30-L73

cole-h:
Seems to be from here: https://github.com/emoon/rust_minifb/blob/8b3c2e9b37d91d21ba63adf15ab2ae44a1578da7/src/native/unix/scalar.cpp

emoon:
Yes this code is done in C in order to have optimizations on even when the Rust code is compiled in debug because they have to be fast

nifker:
So the julia example works now.
The title_cursor basically works now too, I just need to properly implement.
@emoon What about moving that code into an own subcrate and enable optimizations always.

emoon:
Well the code is already written and works so I'm not sure it's needed?

nifker:
Id guess this could then require less unsafe code.

nifker:
This shouldnt be closed.

emoon:
Sure but all minifb almost does is to call into native APIs (even if they are wrapped) so I'm not it will be a big improvement really. I think we can leave them as is and if this needs to be changed later it can be done at that point.

nifker:
Should the content of the window be scaled when the surface gets resized(As I think X11 did it on their own)? Or should it wait for the next update_with_buffer()?

emoon:
If it works fine to scale while the windows is resized that would be great

nifker:
This would definitely work, but I am currently not sure what impact on the performance this would have.

emoon:
Give it a try and see how it works

nifker:
Now on the is_active() method: Is the keyboard focus enough or does pointer focus need to be given too?

cole-h:
Scaling looks like it's working (mostly) fine. However, when running the `julia` and `image` examples, doing whacky resizing (e.g. throwing your pointer around the screen basically) causes the example to crash.

<details><summary>RUST_BACKTRACE=1</summary>

```
[destroyed object]: error 2: error accessing SHM buffer
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: WindowCreate("Event dispatch failed: Os { code: 71, kind: Other, message: \"Protocol error\" }")', src/libcore/result.rs:1188:5
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:84
   3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
             at src/libstd/sys_common/backtrace.rs:61
   4: core::fmt::write
             at src/libcore/fmt/mod.rs:1025
   5: std::io::Write::write_fmt
             at src/libstd/io/mod.rs:1426
   6: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:65
   7: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:50
   8: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:193
   9: std::panicking::default_hook
             at src/libstd/panicking.rs:210
  10: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:471
  11: rust_begin_unwind
             at src/libstd/panicking.rs:375
  12: core::panicking::panic_fmt
             at src/libcore/panicking.rs:84
  13: core::result::unwrap_failed
             at src/libcore/result.rs:1188
  14: core::result::Result<T,E>::unwrap
             at /rustc/5e1a799842ba6ed4a57e91f7ab9435947482f7d8/src/libcore/result.rs:956
  15: minifb::os::unix::wayland::Window::update
             at src/os/unix/wayland.rs:599
  16: minifb::os::unix::wayland::Window::update_with_buffer_stride
             at src/os/unix/wayland.rs:757
  17: minifb::os::unix::Window::update_with_buffer_stride
             at src/os/unix/mod.rs:60
  18: minifb::Window::update_with_buffer
             at ./src/lib.rs:291
  19: image::main
             at examples/image.rs:38
  20: std::rt::lang_start::{{closure}}
             at /rustc/5e1a799842ba6ed4a57e91f7ab9435947482f7d8/src/libstd/rt.rs:67
  21: std::rt::lang_start_internal::{{closure}}
             at src/libstd/rt.rs:52
  22: std::panicking::try::do_call
             at src/libstd/panicking.rs:292
  23: __rust_maybe_catch_panic
             at src/libpanic_unwind/lib.rs:78
  24: std::panicking::try
             at src/libstd/panicking.rs:270
  25: std::panic::catch_unwind
             at src/libstd/panic.rs:394
  26: std::rt::lang_start_internal
             at src/libstd/rt.rs:51
  27: std::rt::lang_start
             at /rustc/5e1a799842ba6ed4a57e91f7ab9435947482f7d8/src/libstd/rt.rs:67
  28: main
  29: __libc_start_main
  30: _start
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```
</details>

nifker:
I have seen the error some days ago as I am not sure what was causing that I was not able to fix it yet.Ill try to investigate on that the next days.

nifker:
@cole-h Fixed with the latest commit.

cole-h:
Confirmed, thanks!

Another issue I notice while doing that crazy resizing: sometimes I see artifacts outside of the window. These artifacts clear after unfocusing the window

image

nifker:
Ive seen this before too, I am trying to fix it currently - it seems to have to do with the surface damaging.

nifker:
Ok Ive just found out it has to do with the buffers being all allocated to one fd - we should instead create a new file for each buffer.

nifker:
Ive created a new branch "bufferpool" which creates the buffers differently, can someone check and see if it works better?
In my case it didnt help remove these artifacts outside the surface. Maybe someone has an idea on how to fix it/finds an issue in my code.

nifker:
Maybe @emersion would be so nice and take a look.

nifker:
Seems like I actually didnt really pushed the bufferpool stuff.
I have now tested a bit around and it seems to persist on older sway versions but neither on weston nor sway-master.
@cole-h Could you test this?

cole-h:
Just built `sway` and `wlroots` master, switched to your `bufferpool` branch, and ran `cargo run --example julia`, and am still seeing the artifacts:

image

nifker:
What about weston?

nifker:
@cole-h Should be fixed now, as someone from the sway IRC gave me a hint.

cole-h:
Confirmed fixed for me. Just for posterity's sake: problem didn't occur on `weston`.

nifker:
So I now assumed that is_active() only counts for the keyboard focus not the mouse focus too(I will push the change soon).
The next and basically last thing is keyboard input. Does someone here have experience with XKB already?

emersion:
>Maybe @emersion would be so nice and take a look.

Sorry, unfortunately I won't have time for this at the moment. Feel free to ask Wayland-related questions on IRC in #wayland though!

nifker:
Seems like most keyboard input stuff works now - maybe some testing!
And I couldnt consider the modifiers so far see: https://docs.rs/xkb/0.2.1/xkb/state/struct.Update.html#method.mask (I dont know from where I need to take the last three arguments)
@emoon does "is_active()" only consider the window focus(as in keyboard), right?

cole-h:
I'm curious if there was a reason you went with the `xkb` crates and decided not to use `wayland-client`'s `wl_keyboard` (e.g. https://github.com/Smithay/wayland-rs/blob/master/wayland-client/examples/simple_window.rs#L128).

---

I think you should use the opposite of what you're currently using for window focus. Right now, when you spawn a window in a tiling window manager like Sway, the window doesn't receive inputs (e.g. `ESC`) until I focus something else and then re-focus that window.

nifker:
@cole-h Oh I didnt know that existed

Thats a bug in wlroots: https://github.com/swaywm/wlroots/issues/1948

cole-h:
I think it might be interesting to research what `winit`/`glutin` have done with regards to focus. Right now, using Alacritty (which uses `glutin`) and opening a new terminal window allows me to type immediately. Maybe they have a special way of handling this.

nifker:
Should to be fixed now.

nifker:
Now were just missing the Unixmenus - I personally have no clue how they should be implemented and how they work(or if there is a standard), but maybe @emoon could tell me.

nifker:
As it seems to me these menus are very custom and are implemented very lightweight on x11, so they dont do much. I would then just use the existing code from the x11 module. But Ill throw my question into here:  Why are these menu functions then even provided if they dont do very much(as there is nothing specific which can be implemented, because it doesnt exist)?

nifker:
Functionality wise Im done with the code. We are just missing some testers for the last part and a code review.

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+            )
+        };
+        tmp_f.write_all(&slice[..]).unwrap();
+        tmp_f.flush().unwrap();
+
+        let xdg_wm_base = global_man.instantiate_exact::<XdgWmBase>(1).map_err(|e| {
+            Error::WindowCreate(format!("Failed retrieving the XdgWmBase: {:?}", e))
+        })?;
+
+        //Ping Pong
```
cole-h: Humorous, but ultimately useless, comment.

```suggestion
        // Reply to ping
```
or
```suggestion
        // Handle ping
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+        xdg_wm_base.quick_assign(|xdg_wm_base, event, _| {
+            use wayland_protocols::xdg_shell::client::xdg_wm_base::Event;
+
+            if let Event::Ping { serial } = event {
+                xdg_wm_base.pong(serial);
+            }
+        });
+
+        let xdg_surface = xdg_wm_base.get_xdg_surface(&surface);
+        let _surface = surface.clone();
```
cole-h: Variables prefixed with an underscore are typically "unused." Since this one is used, opt for a more descriptive name.

```suggestion
        let surface_clone = surface.clone();
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+                _surface.commit();
+            }
+        });
+
+        //Assigns the toplevel role and commit
+        let xdg_toplevel = xdg_surface.get_toplevel();
+        if decoration {
+            use wayland_protocols::unstable::xdg_decoration::v1::client::zxdg_decoration_manager_v1::ZxdgDecorationManagerV1;
+
+            if let Ok(deco_man) = global_man
```
cole-h: Again, more descriptive naming.

```suggestion
            if let Ok(decorations) = global_man
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+                    *xdg_config.borrow_mut() = Some(serial);
+                }
+            });
+        }
+
+        let cursor = wayland_cursor::load_theme(None, 16, &shm);
+        let cursor_surface = comp.create_surface();
+
+        Ok((
+            Self {
```
cole-h: Since it looks like all these variables are used, there's no need to prefix them with an underscore.

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+        }
+
+        self.surface.attach(Some(buf), 0, 0);
+        self.surface
+            .damage(0, 0, i32::max_value(), i32::max_value());
+        self.surface.commit();
+    }
+
+    //Keyboard, Pointer, Touch
+    fn _check_capabilities(
```
cole-h: Considering this appears unused, what is its purpose in the first place?

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+            .map_err(|e| Error::WindowCreate(format!("Roundtrip failed: {:?}", e)))
+            .unwrap();
+
+        let ret = (
+            *keyboard_fl.borrow(),
+            *pointer_fl.borrow(),
+            *touch_fl.borrow(),
+        );
+
+        ret
```
cole-h:
```suggestion
        (*keyboard_fl.borrow(), *pointer_fl.borrow(), *touch_fl.borrow())
```

(or however `rustfmt` decides to format it)

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+    }
+}
+
+struct WaylandInput {
+    kb_events: mpsc::Receiver<wayland_client::protocol::wl_keyboard::Event>,
+    pt_events: mpsc::Receiver<wayland_client::protocol::wl_pointer::Event>,
```
cole-h: Maybe you could import `wayland_client::protocol::wl_{keyboard,pointer}` up to to shorten this to:
```suggestion
    kb_events: mpsc::Receiver<wl_keyboard::Event>,
    pt_events: mpsc::Receiver<wl_pointer::Event>,
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+    input_devs: (Main<WlKeyboard>, Main<WlPointer>),
```
cole-h: Does this need to be a tuple?
```suggestion
    keyboard: Main<WlKeyboard>,
    pointer: Main<WlPointer>,
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+}
+
+impl WaylandInput {
+    fn new(seat: &Main<WlSeat>) -> Self {
+        let (keyboard, pointer, _touch) = (seat.get_keyboard(), seat.get_pointer(), ());
```
cole-h:
```suggestion
        let (keyboard, pointer) = (seat.get_keyboard(), seat.get_pointer());
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+        //requires version 5 for the scroll events
+        let seat = global_man
+            .instantiate_exact::<WlSeat>(5)
+            .map_err(|e| Error::WindowCreate(format!("Failed retrieving the WlSeat: {:?}", e)))?;
+
+        //create the input devices already at this point
+        let input = WaylandInput::new(&seat);
+
+        //retrieve some types from globals
+        let comp = global_man
```
cole-h:
```suggestion
        let compositor = global_man
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+        source_stride: u32,
+        dest_width: u32,
+        dest_height: u32,
+    );
+}
+
+struct SingleBuffer {
+    fd: std::fs::File,
+    //Shm pool and size
+    pool: (Main<WlShmPool>, i32),
```
cole-h: Is there a reason these are a tuple, or could they be split into:
```suggestion
    pool: Main<WlShmPool>,
    pool_size: i32,
```
?

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+        }
+    }
+
+    fn get_pointer(&self) -> &Main<WlPointer> {
+        &self.input_devs.1
+    }
+
+    fn iter_keyboard_events(
+        &self,
+    ) -> std::sync::mpsc::TryIter<wayland_client::protocol::wl_keyboard::Event> {
```
cole-h: Per above:

```suggestion
    ) -> mpsc::TryIter<wl_keyboard::Event> {
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+        self.kb_events.try_iter()
+    }
+
+    fn iter_pointer_events(
+        &self,
+    ) -> std::sync::mpsc::TryIter<wayland_client::protocol::wl_pointer::Event> {
```
cole-h: Per above:
```suggestion
    ) -> mpsc::TryIter<wl_pointer::Event> {
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+    keymap: Option<xkb::keymap::Keymap>,
+    update_rate: UpdateRate,
+    menu_counter: MenuHandle,
+    menus: Vec<UnixMenu>,
+    input: WaylandInput,
+    resizable: bool,
+    //temporary buffer
+    buffer: Vec<u32>,
+    //configure, close
+    toplevel_info: (Rc<RefCell<Option<(i32, i32)>>>, Rc<RefCell<bool>>),
```
cole-h: Clippy complains about this complex type. Maybe you could define a new type, or split it up?

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+            !opts.borderless,
+        )?;
+        if opts.title {
+            dsp.set_title(name);
+        }
+        if !opts.resize {
+            dsp.set_no_resize((width as i32 * scale, height as i32 * scale));
+        }
+
+        //TODO: maybe outsource
```
cole-h: Expand upon this comment, please.

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+};
+
+use std::cell::RefCell;
+use std::os::unix::io::RawFd;
+use std::rc::Rc;
+use std::sync::mpsc;
+
+// We have these in C so we can always have optimize on (-O3) so they
+// run fast in debug build as well. These functions should be seen as
+// "system" functions that just doesn't exist in X11
```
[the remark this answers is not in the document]
nifker: This is borrowed from x11, I just copied the comment too. I could change this but this would be necessary in the x11 file too.

src/os/unix/wayland.rs
nifker: I thought about putting the initialization of "close" and "configure" into an own function.

src/os/unix/wayland.rs
nifker: Firstly I thought it is mandatory to check the capabilities, but then I left it out there. I currently consider removing it.

src/os/unix/wayland.rs
nifker: There's actually none.
Is it possible to take a look at these changes locally instead of "on github"?

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1182 @@
...
+    //temporary buffer
+    buffer: Vec<u32>,
+    //configure, close
+    toplevel_info: (Rc<RefCell<Option<(i32, i32)>>>, Rc<RefCell<bool>>),
+}
+
+impl Window {
+    pub fn new(name: &str, width: usize, height: usize, opts: WindowOptions) -> Result<Self> {
+        let scale: i32 = match opts.scale {
+            //STUB: currently not working in Wayland
```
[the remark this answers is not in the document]
nifker: Ive been told that its not really possible to retrieve the current display and its width and height on Wayland, so I am unsure how to implement this.

src/os/unix/wayland.rs
nifker: Which would you suggest packing into a struct? - Id put the last ones from key_handler into an own struct.

src/os/unix/wayland.rs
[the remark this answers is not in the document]
nifker: winit lets the server do the scaling as they describe here: https://docs.rs/winit/0.22.0/winit/dpi/index.html#how-is-the-scale-factor-calculated

src/os/unix/wayland.rs
nifker: Clippy didnt complain for me - Ill just push my new changes.

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+            use wayland_protocols::xdg_shell::client::xdg_wm_base::Event;
+
+            if let Event::Ping { serial } = event {
+                xdg_wm_base.pong(serial);
+            }
+        });
+
+        let xdg_surface = xdg_wm_base.get_xdg_surface(&surface);
+        let surface_clone = surface.clone();
+        // Handle Ping
```
cole-h:
```suggestion
        // Handle Configure
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+
+            if let Event::Configure { serial } = event {
+                xdg_surface.ack_configure(serial);
+                surface_clone.commit();
+            }
+        });
+
+        // Assigns the toplevel role and commit
+        let xdg_toplevel = xdg_surface.get_toplevel();
+        if decoration {
```
cole-h:
```suggestion
        if decorate {
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+    }
+
+    fn set_no_resize(&self, size: (i32, i32)) {
+        self.toplevel.set_max_size(size.0, size.1);
+        self.toplevel.set_min_size(size.0, size.1);
+    }
+
+    // Sets a specific cursor style
+    fn update_cursor(&mut self, cursor: &str) {
+        let csr = self.cursor.get_cursor(cursor).unwrap();
```
cole-h: I would personally prefer to shadow `cursor`, because `csr` means absolutely nothing to me. Maybe this is what other libraries call it, though, in which case feel free to disregard.

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+        }
+
+        (configure, close)
+    }
+}
+
+struct WaylandInput {
+    kb_events: mpsc::Receiver<wl_keyboard::Event>,
+    pt_events: mpsc::Receiver<wl_pointer::Event>,
+    _keyboard: Main<WlKeyboard>,
```
cole-h:
```suggestion
    keyboard: Main<WlKeyboard>,
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+        let (pt_sender, pt_receiver) = mpsc::sync_channel(1024);
+
+        pointer.quick_assign(move |_, event, _| {
+            pt_sender.send(event).unwrap();
+        });
+
+        Self {
+            kb_events: kb_receiver,
+            pt_events: pt_receiver,
+            _keyboard: keyboard,
```
cole-h:
```suggestion
            keyboard,
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+
+            Scale::X1 => 1,
+            Scale::X2 => 2,
+            Scale::X4 => 4,
+            Scale::X8 => 8,
+            Scale::X16 => 16,
+            Scale::X32 => 32,
+        };
+
+        let (dsp, input) = DisplayInfo::new(
```
cole-h:
```suggestion
        let (display, input) = DisplayInfo::new(
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+
+    pub fn update(&mut self) {
+        self.display
+            .event_queue
+            .dispatch(&mut (), |_, _, _| {})
+            .map_err(|e| Error::WindowCreate(format!("Event dispatch failed: {:?}", e)))
+            .unwrap();
+
+        if let Some(resize) = (*self.toplevel_info.0.borrow_mut()).take() {
+            //Dont try to resize to 0x0
```
cole-h: Missed one ;)

```suggestion
            // Don't try to resize to 0x0
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+                    use wayland_client::protocol::wl_pointer::ButtonState;
+
+                    let st = (state == ButtonState::Pressed) as u8;
+
+                    match button {
+                        //Left
+                        KEY_MOUSE_BTN1 => self.buttons[0] = st,
+                        //Right
+                        KEY_MOUSE_BTN2 => self.buttons[1] = st,
+                        //Middle
```
cole-h:
```suggestion
                        // Left
                        KEY_MOUSE_BTN1 => self.buttons[0] = st,
                        // Right
                        KEY_MOUSE_BTN2 => self.buttons[1] = st,
                        // Middle
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+                    use wayland_client::protocol::wl_pointer::Axis;
+
+                    match axis {
+                        Axis::VerticalScroll => self.scroll_y = value as f32,
+                        Axis::HorizontalScroll => self.scroll_x = value as f32,
+                        _ => {}
+                    }
+                }
+                //Event::Frame {} => {}
+                //Event::AxisSource { axis_source } => {}
```
cole-h: Any reason these were left unimplemented? They should probably be tagged with `FIXME` or `TODO` if you plan to keep them like that for this PR.

```suggestion
                // Event::Frame {} => {}
                // Event::AxisSource { axis_source } => {}
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+                Event::AxisStop { axis, .. } => {
+                    use wayland_client::protocol::wl_pointer::Axis;
+
+                    match axis {
+                        Axis::VerticalScroll => self.scroll_y = 0.,
+                        Axis::HorizontalScroll => self.scroll_x = 0.,
+                        _ => {}
+                    }
+                }
+                //Event::AxisDiscrete { axis, discrete } => {}
```
cole-h: See my question above.

```suggestion
                // Event::AxisDiscrete { axis, discrete } => {}
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+            }
+        }
+
+        self.key_handler.update();
+    }
+
+    fn handle_key(
+        keymap: &xkb::keymap::Keymap,
+        key: u32,
+        state: wayland_client::protocol::wl_keyboard::KeyState,
```
cole-h:
```suggestion
        state: wl_keyboard::KeyState,
```

And then `use wayland_client::protocol::wl_keyboard;` up above.

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+
+                key::apostrophe => Key::Apostrophe,
+                key::grave => Key::Backquote,
+                key::backslash => Key::Backslash,
+                key::comma => Key::Comma,
+                key::equal => Key::Equal,
+                key::bracketleft => Key::LeftBracket,
+                key::minus => Key::Minus,
+                key::period => Key::Period,
+                key::braceright => Key::RightBracket,
```
cole-h: `braceright` and `bracketright` are two different things in `xkb`. Also, it might be nice to have these "sequential" keys be next to each other (e.g. `bracketleft` and `bracketright` right above/below the other).

```suggestion
                key::bracketright => Key::RightBracket,
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+    fn handle_keymap(keymap: KeymapFormat, fd: RawFd, len: u32) -> xkb::keymap::Keymap {
+        use std::io::Read;
+        use std::os::unix::io::FromRawFd;
+
+        match keymap {
+            KeymapFormat::XkbV1 => {
+                use xkb::keymap::Keymap;
+
+                unsafe {
+                    //read in fd content into vec
```
cole-h:
```suggestion
                    // Read fd content into a Vec
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+                    let mut file = std::fs::File::from_raw_fd(fd);
+                    let mut v = Vec::with_capacity(len as usize);
+                    v.set_len(len as usize);
+                    file.read_exact(&mut v).unwrap();
+
+                    let ctx = xkbcommon_sys::xkb_context_new(0);
+                    //create keymap from string
```
cole-h:
```suggestion
                    // Create keymap from String
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+    }
+}
+
+unsafe impl raw_window_handle::HasRawWindowHandle for Window {
+    fn raw_window_handle(&self) -> raw_window_handle::RawWindowHandle {
+        let mut handle = raw_window_handle::unix::WaylandHandle::empty();
+        handle.surface = self.display.surface.as_ref().c_ptr() as *mut _ as *mut c_void;
+        handle.display =
+            self.display.wl_display.clone().detach().as_ref().c_ptr() as *mut _ as *mut c_void;
+        raw_window_handle::RawWindowHandle::Wayland(handle)
```
cole-h:
```suggestion

        raw_window_handle::RawWindowHandle::Wayland(handle)
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+        self.menu_counter.0 += 1;
+        handle
+    }
+
+    pub fn add_menu(&mut self, menu: &Menu) -> MenuHandle {
+        let handle = self.next_menu_handle();
+        let mut menu = menu.internal.clone();
+        menu.handle = handle;
+        self.menus.push(menu);
+        handle
```
cole-h:
```suggestion

        handle
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+    }
+
+    pub fn is_active(&self) -> bool {
+        self.active
+    }
+
+    fn next_menu_handle(&mut self) -> MenuHandle {
+        let handle = self.menu_counter;
+        self.menu_counter.0 += 1;
+        handle
```
cole-h:
```suggestion

        handle
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+
+                if let Event::Configure { width, height, .. } = event {
+                    *configure.borrow_mut() = Some((width, height));
+                } else if let Event::Close = event {
+                    *close.borrow_mut() = true;
+                }
+            });
+        }
+
+        (configure, close)
```
cole-h:
```suggestion
        (resolution, close)
```

Since you named the type `ToplevelResolution`, I think it would make sense to inherit the name here, as well.

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+    // Option because MaybeUninit's get_ref() is nightly-only
+    keymap: Option<xkb::keymap::Keymap>,
+    update_rate: UpdateRate,
+    menu_counter: MenuHandle,
+    menus: Vec<UnixMenu>,
+    input: WaylandInput,
+    resizable: bool,
+    // Temporary buffer
+    buffer: Vec<u32>,
+    // Configure, close
```
cole-h: Just to confirm my suspicions: `ToplevelClosed` means whether or not the window is closed ("minimized"), correct?

```suggestion
    // Resolution, closed
```

src/os/unix/wayland.rs
```diff
@@ -0,0 +1,1117 @@
...
+    toplevel_info: (ToplevelResolution, ToplevelClosed),
+}
+
+impl Window {
+    pub fn new(name: &str, width: usize, height: usize, opts: WindowOptions) -> Result<Self> {
+        let scale: i32 = match opts.scale {
+            // Relies on the fact that this is done by the server
```
cole-h: I think it would be a good idea to link this below, as an additional explanation:
```suggestion
            // Relies on the fact that this is done by the server
            // https://docs.rs/winit/0.22.0/winit/dpi/index.html#how-is-the-scale-factor-calculated
```

src/os/unix/wayland.rs
nifker: Why? "keyboard" is not read afterwards.

src/os/unix/wayland.rs
nifker: Yes whether or not the client wants to close the window.

src/os/unix/wayland.rs
cole-h: Ah, I was conflating public API and private API. If it's not used, then yes this is fine -- sorry.

nifker:
It would be nice if you a new PR for this.

nifker:
@cole-h Have you come further on the PR?

cole-h:
Ah, I totally forgot! Sorry; with all the craziness this past week (classes getting cancelled, coronavirus, etc) it slipped my mind. I'll get to work on it.

nifker:
I now removed most unwrapping here and there, but there's some remaining because update and update_cursor don't return a Result.

nifker:
@emoon What do you think about the current status of this? Are we missing something, or similar?

emoon:
if everything is in place we can merge it but current the CI is failing. I'm not sure if it's related to your changes or not?

nifker:
@emoon Its because theres no xkbcommon in the CI. I dont know how to add it there.

emoon:
Isn't it possible to use something like `x11-dl` style that uses dynamic loading and doesn't require a lib to be required?

nifker:
I dont think it would be useful for a wayland impl to be relying on x11 dependencies. Ill take a look if I can add the dependency somehow.

emoon:
I don't suggest to depend on x11 dependencies but load the lib in a dynamic fashion the way it works on x11

nifker:
I sadly dont know how to do it.

cole-h:
I think that relies more on the upstream's implementation, rather than something that can be done here. The Ubuntu CI just needs to add the `libxkbcommon` (or `-dev`) package to its dependencies list. Might also need some Wayland deps, but one thing at a time.

emoon:
Sure but this also adds an extra dependency for minifb that has to be added outside of cargo that wasn't required before.

nifker:
Why is this an issue to add a dependency to the CI?

emoon:
It's not a problem adding the dependency on the CI but it means that users of minifb will run into the same issue as well no?

cole-h:
Would that be solved by feature-gating the Wayland and X11 implementations? I say "and" 1) for people who have no X11 libraries installed and want to develop for Wayland-only; and 2) because the X11 crates would be unnecessary in this scenario of only developing for Wayland.

emoon:
The way I understand it is that you can only detect Wayland and x11 in run-time (not at compile time)

cole-h:
No need to detect: make the consumer specify whether or not they want Wayland, X11, or both to be supported.

It's not possible to dynamically detect at compile-time, yes, but by using Cargo features (see for example ripgrep, which [conditionally enables PCRE2 support](https://github.com/BurntSushi/ripgrep/blob/master/Cargo.toml#L82) based on a feature flag), one could force this decision on the downstream users.

emoon:
Yes that would be an acceptable compromise I think.

cole-h:
For inspiration, here's the commit introducing the `x11`/`wayland` feature flags into Alacritty: https://github.com/alacritty/alacritty/commit/4000ec04d89b9bd77995960d2a9da8fcad55e003. Not as substantial as this might become, but it should give @nifker a rough idea of how to approach this.

nifker:
Can someone test the feature gate?

cole-h:
Tested with the `julia` example, and both `x11` and `wayland` features appear to work, but you should also feature-gate the platform-specific dependencies.

Similar to my earlier example: if someone wants to develop solely for X11, they'd still be pulling in `wayland-client` and friends as dependencies. Probably also want to feature-gate the `wayland` and `x11` modules (and related methods/structs/enum variants/etc).

Also, only tangentially related to this PR, but maybe it would be better to group the Unix dependencies under a single, common header as done [in one of Alacritty's `Cargo.toml`s](https://github.com/alacritty/alacritty/blob/master/alacritty_terminal/Cargo.toml#L29). This would reduce the ~9 duplicate entries... (Keep in mind that you can use similar syntax to -- if not the same as -- the `cfg` attribute macro.)

nifker:
This should do it.

src/os/unix/mod.rs
```diff
@@ -8,8 +8,8 @@
 // turn off a gazillion warnings about X keysym names
 #![allow(non_upper_case_globals)]
 
-//mod wayland;
 mod key_mapping;
+mod wayland;
 mod x11;
```
cole-h: Guard these.

src/os/unix/mod.rs
```diff
@@ -19,26 +19,43 @@ pub use x11::Menu;
 
 use std::os::raw;
 
-//Differentiate between Wayland and X11 at run-time
+// Differentiate between Wayland and X11 at run-time
 pub enum Window {
     X11(x11::Window),
-    Wayland(()), //WlWindow
+    Wayland(wayland::Window),
```
cole-h: And these.

Should probably also make this nonexhaustive by adding something like:

```rust
#[doc(hidden)]
__Nonexhaustive,
```

src/os/unix/mod.rs
```diff
@@ -19,26 +19,43 @@ pub use x11::Menu;
...
 }
```
cole-h: Basically everything in this file that refers to one or the other needs to be feature-gated.

nifker:
Isnt there a simpler way than insert hundred conditional compilation conditions?

cole-h:
Splitting the file up further and conditional imports, maybe. Might be useful to see what `winit`/`glutin` do here.

src/os/unix/mod.rs
nifker: Why do you think this is important?

cole-h: My reasoning was: to force consumers to handle all cases. Now, I can see that it isn't important.

build.rs
```diff
@@ -2,6 +2,16 @@ use std::env;
 extern crate cc;
 
 fn main() {
+    if cfg!(not(any(
+        target_os = "macos",
+        target_os = "windows",
+        target_os = "redox"
+    ))) {
+        if cfg!(not(any(feature = "wayland", feature = "x11"))) {
```
cole-h:
```suggestion
    ))) && cfg!(not(any(feature = "wayland", feature = "x11"))) {
```

build.rs
```diff
@@ -2,6 +2,15 @@ use std::env;
 extern crate cc;
 
 fn main() {
+    if cfg!(not(any(
+        target_os = "macos",
+        target_os = "windows",
+        target_os = "redox"
+    ))) && cfg!(not(any(feature = "wayland", feature = "x11"))) {
+            panic!("At least one of the x11 or wayland features must be enabled");
+        }
```
cole-h:
```suggestion
```

nifker:
So we are now just missing the dependencies(xkbcommon and libwaylandcursor)

nifker:
@emoon Do we (except for the deps) miss something?

emoon:
I guess not. Can you add the missing deps to the CI script also to make sure that the CI passes and perhaps add CI step to build with x11 and wayland separate to make sure both of them work?

nifker:
@emoon I guess your talking about this script: https://github.com/emoon/rust_minifb/blob/master/.github/workflows/ci.yml
Do you have any docs about it? As I cant instantly see where I need to add the dependencies.

emoon:
Yeah docs is over here https://help.github.com/en/actions 

I think you will need to do something like this

```
    - name: Install deps
      run: sudo apt-get install libasound2-dev mesa-common-dev

**Comment by emoon** [comment]
You should be able to do something like this

```
 - if: matrix.os == 'ubuntu-latest'
      run: sudo apt install ...
```

It would be good to have it set up in such way we have one test for wayland compile and one for x11 (as x11 shouldn't need the extra dependency)

**Comment by nifker** [comment]
@emoon are there any examples in the docs or a list of possible keywords and syntax, etc? Cant find any, so far.

**Comment by nifker** [comment]
Ok now we have some weird error in the CI.

File: Cargo.toml [hunk 43]
```diff
@@ -39,65 +39,19 @@ features = [
...
-x11-dl = "2.18.3"
-wayland-client = "0.25"
-wayland-protocols = { version = "0.25", features = ["client"] }
-tempfile = "3.1.0"
+[features]
+default = ["wayland", "x11"]
+x11 = ["x11-dl"]
+wayland = ["wayland-client", "wayland-protocols", "wayland-cursor", "tempfile", "xkb", "xkbcommon-sys"]
+
+[target.'cfg(unix)'.dependencies]
```
cole-h, line 47:
```suggestion
[target.'cfg(not(target_os = "macos"))'.dependencies]
```

Pretty sure `unix` also counts macOS.

nifker: This will include redox, so we need to change it

cole-h: Updated.

.github/workflows/ci.yml
```diff
@@ -28,6 +28,8 @@ jobs:
         os: [ubuntu-latest, windows-latest, macOS-latest]
     runs-on: ${{ matrix.os }}
     steps:
+      - if: matrix.os == 'ubuntu-latest'
+        run: sudo apt install libwayland-cursor0 libxkbcommon-dev
```
cole-h, line 32:
```suggestion
        run: sudo apt install libwayland-cursor0 libxkbcommon-dev libwayland-dev
```

Should fix CI once and for all.

emoon:
Merged! Thanks for all the work on this :) Open a new PR if you want to do some more fixes for it as this one is pretty big already. I will try to release a new version today

Issue #150, opened 2020-03-17
There is no `get_keys_released` method on the `Window` struct


Melvillian:
Background: I am writing an emulator that uses `rust_minifb` for window state and keyboard input. The emulator has its own array to store the state of pressed keys. I want to be able to update the emulator's keyboard using  `minifb`'s `Window` struct.

Problem: I cannot detect in bulk which keys have been released in the last window `update`, and so will need to write my own code which calculates, given the return value of `window.get_keys_pressed`, which keys are not included in that `&Option<&Vec<Key>>` and sets those to no longer be pressed in the emulator. This feels like a common requirement for `minifb` users.

Solution: It would be more ergonomic for my emulator to call a `window.get_keys_released` and then update it's state using only those keys which changed from pressed to released. This would also nicely mirror `get_keys_pressed` which already exists.

I wanted to check that I'm not missing something before I go and submit a PR for this. If this seems like a good idea to you please just give me the thumbs up and a PR will be inbound shortly.

emoon:
Hey, a PR for this sounds good 👍

Melvillian:
PR at https://github.com/emoon/rust_minifb/pull/151

PR #151, opened 2020-03-17
add get_keys_released method to Window

Melvillian:
See #150 

Adds a method `get_keys_released` to all `Window` implementations and extends the noise example to call it. This is the inverse of `get_keys_pressed` method in that it will fill the returned `Vec` with `Keys` that were released in the last update, rather than pressed.


emoon:
Thanks!

PR #152, opened 2020-03-18
Change Error to use custom fmt::Debug messages

leniseev:
std::error::Error::description is deprecated, so I changed the
fmt::Display implementation to rely on fmt::Debug implementation, which,
in turn, return human-readable messages like description was doing


emoon:
Great! Thanks :)

PR #154, opened 2020-03-30
Added Topmost/Always on Top functionality to MacOS 

phillvancejr:
This functionality is useful for utility apps that he developer wants to float above other windows

WindowOptions now has a topmost property denoting whether the window should be the topmost window.  The mfb_topmost function will set the NSWindow level property to NSFloatingWindowLevel if WindowOptions::topmost is true, otherwise it will be 0 (normal).


emoon:
Thanks!

PR #156, opened 2020-03-31
Removing some legacy Debug information

phillvancejr:
Removed NSLog(@"Names %@", [g_library functionNames]); debug call from src/native/macosx/MacMiniFB.m


emoon:
Thanks!

PR #163, opened 2020-04-05
[Testing] Implement is_active() on windows and x11

nifker:
X11 seems to work just fine.
Would be nice if someone can test this on windows.


emoon:
Thanks!

Issue #133, opened 2020-01-20
Transparent Window


nifker:
Should transparency be allowed for bordered windows?
TBD
Which platforms support window transparency?
TBD

- [x] Windows
- [x] OSX
- [x] Linux(X11/Wayland)
- [x] Redox

How should it be enabled?
A: Set a flag in struct WindowOptions.

nifker:
@FloVanGH Does Redox support window transparency?

nifker:
We might let this open as its not implemented on Windows and OSX

emoon:
Do you plan to implement support for Windows and macOS also?

nifker:
@emoon I will definitely try to implement it on windows but not on osx because I dont have a device to test on it.

nifker:
I will link to some working implementation for alpha transparency on OSX as well as the Windows one: 
https://github.com/godotengine/godot/pull/14622/commits/ddae098deec89d67228a74623ebe7f5db39f425b#diff-65f347288a42e15c028ce4f494698fa1R2137-R2142

emoon:
Alright. That code uses OpenGL on macOS which is deprecated. I use Metal on macOS which is the supported API but I guess it should be possible to do the same thing.

rupansh:
@nifker 
Manually using winapi isn't working either
```
    let wind = Window::new("", w as usize, h as usize, WindowOptions {
        title: false,
        borderless: true,
        ..WindowOptions::default()
    }).unwrap();
    match wind.raw_window_handle() {
        RawWindowHandle::Windows(wnd) => {
            let hwnd = wnd.hwnd as HWND;
            unsafe { SetWindowLongA(hwnd, GWL_EXSTYLE, (WS_EX_LAYERED | WS_EX_TRANSPARENT) as i32) };
            unsafe { SetLayeredWindowAttributes(hwnd, 0, 255, LWA_ALPHA)};
            let mut margs = MARGINS {
                cxLeftWidth: -1,
                cxRightWidth: -1,
                cyTopHeight: -1,
                cyBottomHeight: -1,
            };
            unsafe { DwmExtendFrameIntoClientArea(hwnd, &mut margs) };
            unsafe { UpdateWindow(hwnd) };
            unsafe { SetWindowPos(hwnd, HWND_TOPMOST, l, t, w as i32, h as i32, SWP_SHOWWINDOW) };
        }
        _ => {}
    }
```

However the same code works fine with sfml window and Style::NONE

PR #164, opened 2020-04-07
Transparency

nifker:
This will partially resolve https://github.com/emoon/rust_minifb/issues/133
Implemented:
- X11
- Wayland
- Redox(couldnt test so far)

### DIFF
--- a/examples/transparent.rs
+++ b/examples/transparent.rs
@@ -1,0 +1,28 @@
+extern crate minifb;
+
+use minifb::{Key, ScaleMode, Window, WindowOptions};
+
+fn main() {
+    // Allocate the output buffer.
+    let buf = vec![0x00AAFF33; 320*480];
+
+    let mut window = Window::new(
+        "Press ESC to exit",
+        320 as usize,
+        480 as usize,
+        WindowOptions {
+            resize: true,
+            scale_mode: ScaleMode::Center,
+			borderless: true,
+			transparency: true,
+			..WindowOptions::default()
+        },
+    )
+    .expect("Unable to open Window");
+
+    while window.is_open() && !window.is_key_down(Key::Escape) {
+        window
+            .update_with_buffer(&buf, 320, 480)
+            .unwrap();
+    }
+}
--- a/examples/transparent.rs
+++ b/examples/transparent.rs
@@ -4,25 +4,23 @@ use minifb::{Key, ScaleMode, Window, WindowOptions};
 
 fn main() {
     // Allocate the output buffer.
-    let buf = vec![0x00AAFF33; 320*480];
+    let buf = vec![0x00AAFF33; 320 * 480];
 
     let mut window = Window::new(
         "Press ESC to exit",
-        320 as usize,
-        480 as usize,
+        320,
+        480,
         WindowOptions {
             resize: true,
             scale_mode: ScaleMode::Center,
-			borderless: true,
-			transparency: true,
-			..WindowOptions::default()
+            borderless: true,
+            transparency: true,
+            ..WindowOptions::default()
         },
     )
     .expect("Unable to open Window");
 
     while window.is_open() && !window.is_key_down(Key::Escape) {
-        window
-            .update_with_buffer(&buf, 320, 480)
-            .unwrap();
+        window.update_with_buffer(&buf, 320, 480).unwrap();
     }
 }
--- a/examples/transparent.rs
+++ b/examples/transparent.rs
@@ -1,5 +1,3 @@
-extern crate minifb;
-
 use minifb::{Key, ScaleMode, Window, WindowOptions};
 
 fn main() {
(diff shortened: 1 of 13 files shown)


examples/transparent.rs
```diff
@@ -0,0 +1,28 @@
...
+
+use minifb::{Key, ScaleMode, Window, WindowOptions};
+
+fn main() {
+    // Allocate the output buffer.
+    let buf = vec![0x00AAFF33; 320*480];
+
+    let mut window = Window::new(
+        "Press ESC to exit",
+        320 as usize,
```
emoon, line 11: shouldn't need to say `value as usize` here

nifker:
@cole-h Are you interested to help here?

src/lib.rs
```diff
@@ -187,6 +187,9 @@ pub struct WindowOptions {
     pub scale_mode: ScaleMode,
     /// Should the window be the topmost window (default: false)
     pub topmost: bool,
+    /// Specifies whether or not the window is allowed to draw transparent pixels(default: false)
```
cole-h, line 190:
```suggestion
    /// Specifies whether or not the window is allowed to draw transparent pixels (default: false)
```

src/os/unix/x11.rs
```diff
@@ -90,8 +90,8 @@ struct DisplayInfo {
 }
 
 impl DisplayInfo {
-    fn new() -> Result<DisplayInfo> {
-        let mut display = Self::setup()?;
+    fn new(transparency: bool) -> Result<DisplayInfo> {
```
cole-h, line 93: I don't really like that this takes `transparency` as an argument, but I don't have any suggestions to improve this...

src/os/unix/x11.rs
```diff
@@ -115,10 +115,30 @@ impl DisplayInfo {
...
+                (lib.XMatchVisualInfo)(
+                    display,
+                    (lib.XDefaultScreen)(display),
+                    32,
+                    xlib::TrueColor,
+                    &mut vinfo as *mut _,
+                );
+                depth = vinfo.depth;
+                visual = vinfo.visual;
+                screen = vinfo.screen;
```
cole-h, line 134: Just a style nit, but would be nice to follow declaration order above and assignment order below:
```suggestion
                screen = vinfo.screen;
                visual = vinfo.visual;
                depth = vinfo.depth;
```

examples/transparent.rs
```diff
@@ -0,0 +1,26 @@
+extern crate minifb;
```
cole-h, line 1: I thought Rust 2018 made this unnecessary.

src/os/unix/x11.rs
nifker: The x11 one currently doesnt even work, when you try the transparency example: it produces a BadMatch on XPutImage.

nifker: Im currently trying to find out where the problem is :/

emoon:
Question:
With transparency I assume you can set how visible a window should be? Maybe it would be better with an `Option<f32>` instead of a bool where the `Some(0.5)` would be 50% transparency, `Some(0.75)` 75% etc

nifker:
@emoon This would be possible to implement in theory, but needs to be simulated. The plan here is to allow alpha transparency for windows, where the dev can draw (semi)transparent RGBA data to the window.

emoon:
Alright. Then I guess a bool makes more sense

nifker:
X11 is now working too(but requires an compositor).
@emoon Can you research on how to do this on OSX and implement the code for it?

src/os/unix/x11.rs
nifker: I actually fixed it, open to improvements now.

nifker:
Ok the windows one will take longer, because windows requires pre-multiplied colours https://github.com/godotengine/godot/pull/14622/commits/ddae098deec89d67228a74623ebe7f5db39f425b#diff-33050a8c8ed21f4806d591c06f251d48R1720-R1740

nifker:
Seems like I am not able to currently implement transparency so far(as of time and experience with windows and osx).
@emoon Would it be possible to review this and merge the linux/bsd/etc. implementation and the windows/osx implementation would come later?

emoon:
Yes that is fine. Should I merge these changes now?

nifker:
@emoon Next time I push Id be ready.

src/lib.rs
```diff
@@ -220,6 +224,12 @@ impl Window {
     ///  .expect("Unable to open Window");
     /// ```
     pub fn new(name: &str, width: usize, height: usize, opts: WindowOptions) -> Result<Window> {
+        if opts.transparency {
+            assert!(
```
nifker, line 228: @emoon Should we assert here or return an Err value?

emoon: Return an error code. `assert!` will only be present in debug config.

nifker: Ok Im done here

emoon:
ok, good to merge now?

nifker:
Yes

PR #174, opened 2020-04-27
Add method for updating the buffer only

christoph-heiss:
Hi,
I basically needed a way to update the framebuffer independently of input due to blocking and expense of updating the input state.

I had this branch laying around for some time now, I rebased & updated the changes so the branch can be merged, before it completely code-rots.  Although very specialized, I'd like to get this upstream.

Tested only on Linux/X11, but all the code is basically a copy of `update_with_buffer_stride` with any input updates ripped out.

I'm open for suggestions and be willing to further work on this if needed in order to get this merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ fn main() {
         }
 
         // We unwrap here as we want this code to exit if it fails. Real applications may want to handle this in a different way
+        window.update();
         window
             .update_with_buffer(&buffer, WIDTH, HEIGHT)
             .unwrap();
--- a/examples/image.rs
+++ b/examples/image.rs
@@ -32,8 +32,9 @@ fn main() {
     .expect("Unable to open Window");
 
     while window.is_open() && !window.is_key_down(Key::Escape) {
+        window.update();
         window
-            .update_with_buffer(&u32_buffer, info.width as usize, info.height as usize)
+            .update_buffer(&u32_buffer, info.width as usize, info.height as usize)
             .unwrap();
     }
 }
--- a/examples/julia.rs
+++ b/examples/julia.rs
@@ -61,7 +61,8 @@ fn main() {
         angle += 0.1;
 
         // We unwrap here as we want this code to exit if it fails
-        window.update_with_buffer(&buffer, WIDTH, HEIGHT).unwrap();
+        window.update();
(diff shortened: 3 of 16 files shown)


nifker:
I think it might be senseful then to remove the update_buffer_with_stride function. So that we only have a function which updates input and one which updates the framebuffer(in the public API).

nifker:
@christoph-heiss I think it might be senseful then to remove the update_buffer_with_stride function. So that we only have a function which updates input and one which updates the framebuffer(in the public API).

emoon:
I agree with @nifker here. That would make it a bit cleaner

christoph-heiss:
As suggested, I've removed the `.update_buffer_with_stride()` completely.
I then also renamed the remaining method to simply `.update_buffer()` to better represent the new API.

The second commit just adapts all examples to the new `Window` API.

As always, only tested on Linux/X11 and RFC'ing on this.

nifker:
@christoph-heiss Seems fine to me. But Wayland is struggling a bit here - Ill take a look at it.

nifker:
@christoph-heiss Ok we are using single files on Wayland for a framebuffer so it would be nice if you include a note for Wayland that "when the buffer is too often updated without using update() the application will crash because too many files have been opened".

nifker:
I think there is no compositor which holds 20 framebuffers at the same time.
I would add this in line '1104' in 'wayland.rs':
"assert!(self.display.buf_pool.pool.len() < 20, "Leaking file descriptors");"

christoph-heiss:
@nifker TBH, that sounds kind of like a hack to me. I'd be willing to work on a proper solution for this and can set me up a wayland-environment for this, I guess.

This sounds like the compositor needs to explicitly "consume" the frames in-flight we write out to the fd (don't really know the proper wayland-way words, never worked with it before), and this behaviour must be kind of hidden(?) somewhere in `.update()`.

I'll see what I can do when I get around that, so the PR can stay open as-is for now.

nifker:
Not sure if I can figure something better out, as calling update() will be mandatory anyway, so the developer just needs to make sure to call it once in a while.

emoon:
Closing due to no progress, if this is still of interest we can re-open

PR #180, opened 2020-05-04
Delete old menu system sketch code from 4 years ago

nelsonjchen:
It's commented out and successor menu code is in.


emoon:
Thanks!

PR #183, opened 2020-05-19
Remove usage of the time crate.

### LINKED ISSUE #124: Remote usage of time crate
Looking over the code it only pretty much uses `precise_time_s()` and I think we can just use the one from the stdlib instead and reduce one dependency.

ArifRoktim:
Should fix #124.


emoon:
Thanks!

PR #184, opened 2020-05-19
Use xkbcommon's function to convert a keysym to a unicode point

nifker:
Removed key_mapping.rs => less code.


nifker:
Im unsure if the code_point range check is neccessary or not. Ill take a look if it is.

emoon:
Thanks!

nifker:
@emoon Can you apply the change in my x11 branch too? (forgot to add the deps for x11)

nifker:
@emoon this change could really be important.

emoon:
Can you open a new PR for the fix?

Issue #112, opened 2019-12-03
Physical Keys


nifker:
The ability to check for non-ascii keys/layouts independent would be nice in some cases.

nifker:
@emoon I took a loser look at this now and thought of realizing this - do you think we should extend the KeyHandler struct or rather create a similiar PhysicalKeyHandler struct?

nifker:
@emoon Ive just implemented it for X11 and Wayland in my fork if you would like to take a look. Would like to receive feedback about it!

nifker:
My current idea is that I use a seperate physical key handler and a virtual key handler.
But I think about only using a physical key handler and convert them to  virtual key codes on the fly using lookup tables.

nifker:
I would currently say that this is easier to implement when minifb uses an event loop like winit and SDL2 does.

emoon:
Context from #430: the Wayland backend now derives `Key` from level 0 of the keymap and caches the mapping per keycode across press/release, so `Key` is independent of shift level and of the active layout group.

It is still not script-independent. Level 0 of a non-Latin layout is a non-Latin keysym, so those keys map to nothing — on AZERTY the `1` key is `ampersand` at level 0 and is dropped, and a Cyrillic layout drops every letter. X11 has the same limitation via `XLookupKeysym(&ev.key, 0)`.

Fixing that properly means mapping the raw evdev scancode to `Key` and using xkb only for character input, which is what this issue is asking for.

PR #185, opened 2020-05-21
[WIP] Physical Keys

nifker:
This will close https://github.com/emoon/rust_minifb/issues/112.


nifker:
@emoon would be interested in feedback.

emoon:
I think the code looks good.

nifker:
I am unsure if it is better to convert the system-dependent keycodes to our abstracted scancodes, or just expose the pressed keys as system-dependent u32. Which one could convert themself using a library.

emoon:
Sorry for the late update. I'm not sure either what is best here

emoon:
Closing due to progress for a while. We can re-open this if this is still going to be done.

PR #197, opened 2020-07-09
Rename `unix`es to `posix`

### LINKED ISSUE #195: Linux is not Unix
It's a semi-common convention to call systems that could be either Unix or Linux "*nix systems". BSD is a type of Unix, but Linux is its own thing.

When you use `unix`, you seem to mean both Linux and BSD (which are both *nix systems), but I think it could be more correct to say `nix` instead. Or perhaps `posix`. Many *nix OSes aren't strictly POSIX compliant (macOS is), but almost all of them are POSIX-like. `posix` is the more technical one, while `nix` is more informal.

Honestly, I would go for `posix` for build systems and the like but either one would work. And both are better than saying that Linux is Unix! :P

LoganDark:
Attempts to fix #195 by changing the terminology.

_However_, this set of changes is not yet complete. I need feedback.

References to Unix that include Linux persist in many places, such as `get_unix_menus` and `UnixMenu`. Since it's not very polite to remove/rename these without warning, I would like to deprecate `get_unix_menus` and add `get_posix_menus`, which would do the exact same thing (for now). `get_unix_menus` would simply be an alias to `get_posix_menus`, for API compatibility.

Once version 1.0.0 hits, `get_unix_menus` would be removed and only `get_posix_menus` would be left. `UnixMenu` would be renamed to `PosixMenu`.

Let me know if you have any problems with this strategy.


LoganDark:
Also I think there's something wrong with your CI. Your lint check fails to run rustfmt successfully.

Also the ubuntu check is wrong... I did fix this occurrence
image

LoganDark:
Apparently I still saw that in the diff even though it wasn't recorded in the commit, nevermind about the build.rs thing

emoon:
One could use
```
#[deprecated(
    since = "0.2.1",
    note = "Please use the bar function instead"
)]
```

LoganDark:
So you're okay with the deprecation strategy?

emoon:
Yes. People will still be able to call the old and new name and as you say we can remove the `unix` name when 1.0 hits

LoganDark:
Let me know what you think.

emoon:
Doc fixes looks good. I was also thinking about the `UnixMenu` struct but I think it's best to change it later at 1.0 as you wrote.

LoganDark:
Ready for merge.

emoon:
Thanks!

LoganDark:
You're welcome!

PR #222, opened 2020-10-23
Fix Key::RightBracket not working on POSIX

MightyPork:
There was a typo in the mappings, right bracket doesn't work.


emoon:
Thanks!

PR #223, opened 2020-11-05
Windows: Fix mouse positions not honoring MouseMode

sndels:
Mouse positions went past the right side and bottom of the window even with discard and clamp mouse modes. This fixed things with the mouse example and seems to match what other backends do.

On a side note, the TODO-comments look stale as I didn't see any odd behavior with resizing after the fix.


emoon:
Thanks!

PR #227, opened 2020-12-06
Add linux camera demo

sigmaSd:


sigmaSd:
I opened a PR here https://github.com/raymanfx/libv4l-rs/pull/18 instead since this is linux only I think its more appropriate.

PR #230, opened 2020-12-09
Fix: winuser | DestroyWindow

sinclairzx81:
This PR updates the Windows winuser `Drop` implementation to use `winuser::DestroyWindow(...)` instead of `winuser::CloseWindow(...)`. According to the [docs](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-closewindow), `CloseWindow()` "Minimizes (but does not destroy) the specified window.".  This should allow the Window to be properly destroyed once it falls out of scope.

I have also added a small example `cargo run --example drop` that tests the behavior. Feel free to remove this if not needed. Have left it there in case it was helpful to quickly test this behavior across other OS.

### DIFF
--- a/examples/drop.rs
+++ b/examples/drop.rs
@@ -1,0 +1,30 @@
+use std::thread;
+use std::time::{Duration, Instant};
+use minifb::{ Key, ScaleMode, Window, WindowOptions };
+
+const WIDTH:  usize = 640 / 2;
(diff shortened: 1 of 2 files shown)


emoon:
Thanks! I think I will remove the example as it isn't really needed.

PR #232, opened 2020-12-30
Make Debug for Error include the sub-error.

CoffmanLevi:
The old implementation of Debug for Error would only print what type of error it was. Not what caused the error. Whereas, Display would print the type of error and what caused the error. This seemed backwards to what it should be, so I flipped them.


emoon:
Thanks!

PR #233, opened 2021-01-09
Update xkb_keymap reference

brokenloop:
The xkbcommon_sys crate seems to have been updated today, and now the minifb project won't compile. 

We can fix this by updating XKB_KEYMAP_FORMAT_TEXT_v1 to XKB_KEYMAP_FORMAT_TEXT_V1.

### DIFF
--- a/src/os/posix/wayland.rs
+++ b/src/os/posix/wayland.rs
@@ -1055,7 +1055,7 @@ impl Window {
                     let kb_map_ptr = xkbcommon_sys::xkb_keymap_new_from_string(
                         ctx,
                         v.as_ptr() as *const _ as *const std::os::raw::c_char,
-                        xkbcommon_sys::xkb_keymap_format::XKB_KEYMAP_FORMAT_TEXT_v1,
+                        xkbcommon_sys::xkb_keymap_format::XKB_KEYMAP_FORMAT_TEXT_V1,
                         0,
                     );
 


emoon:
Thanks!

brokenloop:
No worries! Thanks for the great project.

emoon:
So when I try to build locally now I get a compile error due to this change. Is there some cargo dependency that needs to be updated as well?

khyperia:
> So when I try to build locally now I get a compile error due to this change. Is there some cargo dependency that needs to be updated as well?

The issue is that `xkbcommon_sys` made a breaking change without the appropriate semver bump, which can cause all sorts of confusing situations.

- `xkbcommon_sys` `0.7.4` has lowercase `XKB_KEYMAP_FORMAT_TEXT_v1`
- `xkbcommon_sys` `0.7.5` has uppercase `XKB_KEYMAP_FORMAT_TEXT_V1`

Before this change, you must build with `xkbcommon_sys<=0.7.4`. After this change, you must build with `xkbcommon_sys>=0.7.5`

emoon:
I see. So I guess it's best to depend on the exact correct version then?

khyperia:
Yeah, changing the current version to require `0.7.5` is I think the right path forward. I think old versions of `minifb` are just broken and can't be fixed, since a `cargo update` will bump `xkbcommon_sys` to `0.7.5`, but oh well I guess (hopefully the same `cargo update` will bump `minifb` to the latest version with this patch as well).

eddyb:
Did anyone get in touch with `xkbcommon_sys` maintainers? IMO they should yank semver-violating releases and bump the minor version instead.

PR #235, opened 2021-02-10
windows: use c_void from winapi

xobs:
The `SelectObject()` call accepts a `HGDIOBJ` which is typecast to a
`std::os::raw::c_void` when building using libstd, and is its own thing
when building with std disabled:

winapi/0.3.9/src/winapi/lib.rs.html:

```rust
    #[cfg(feature = "std")]
    pub use std::os::raw::c_void;
    #[cfg(not(feature = "std"))]
    pub enum c_void {}
```

This patch uses the winapi c_void as the typecast as necessary.


emoon:
Thanks!

PR #238, opened 2021-02-19
Fix wayland missing cursor.

dc740:
This fixes the problem with the window automatically closing itself when
the mouse goes over it.


nifker:
LGTM - but cant check it myself atm.

emoon:
Thanks!

PR #239, opened 2021-02-28
Wasm websys wip

dc740:
This is a possible implementation, based on @tversteeg code. For the web part to work you need to have an executor and use spawn_local. This implementation does not force any executor, in fact, it does not even care what you use.

I still need to work on an example, and test it. It compiles, and all that really matters on this is to always keep a reference to the window, so the buffer does not go out of scope (which would trigger this: https://github.com/rustwasm/wasm-bindgen/issues/2445 )
Compile using:
```   
cargo build --target=wasm32-unknown-unknown --verbose --no-default-features --features web

### CODE REVIEW

File: src/os/wasm/mod.rs [hunk 1]
```diff
@@ -0,0 +1,321 @@
...
+        
+        /*buffer_helper::check_buffer_size(
+            self.width as usize,
+            self.height as usize,
+            self.window_scale,
+            buffer,
+        )?;*/
+        let mut data = u32_as_u8(buffer);
+
+        let data = ImageData::new_with_u8_clamped_array_and_sh(Clamped(&mut data), self.width, self.height).unwrap();
```
dc740, line 160: I don't really know if this will work or not, but it's worth a try. I need to test it.

Cargo.toml
```diff
@@ -27,6 +27,9 @@ cc = "1.0"
 
 [dependencies]
 raw-window-handle = "0.3.3"
+cast = "0.2"
+time = "0.1.34"
+instant = { version = "0.1.9", features = [ "web-sys", "now" ] }
```
emoon, line 33: Are these required as general dependencies?

dc740: no, I have to clean up this. I don't know why time and cast are here at all. I know I need instant for the web part, but definitely not a general dependency. I'll move it to the web dependencies later.

src/os/wasm/mod.rs
dc740: it worked

Cargo.toml
dc740: A little late, but I removed the dependency from the global section and it's now only required when building the web part

dc740:
Hi, I put some extra work the last couple of days on this:
- Cleaned up the dependencies to make sure we only include the ones we need on the web compilation step. Removed any other dependency from the rest of the work.
- Squashed the huge list of commits into only one.
- Worked on a multi platform demo that can be compiled to desktop AND web: https://github.com/dc740/minifb-async-examples

This is usable now. Getting minifb on the web requires some extra work, and I can't provide a simple example to do "cargo run minifb_web", that's why I had to add the example in other repository.

I kept it as small as possible, and after some research I don't think there is any other way to implement it. There is no way to keep the simple application loop that the desktop uses, but after this PR it's possible to finally run a (mostly canon) minifb application on the web using webassembly and some auto generated javascript glue code as shown in the demo.

I was hoping this PR could be merged so other people can start making improvements on it. Let me know if you think that's possible, or which things you think are blocking this PR.

emoon:
Merged! Thanks for your work on this :)

PR #243, opened 2021-03-13
key_handler: add a callback for key events

xobs:
Add a callback for key events. This includes complete events such as
Shift, not just text events.

A default handler is provided for backwards compatibility.


emoon:
Thanks!

Issue #245, opened 2021-04-02
Incorrect size of non-decorated windows on Windows


swismer:
When creating a non-decorated window on Windows, the size of the window is too large, because the calculation always assumes decoration. This issue is not obvious as the oversize area seems to be transparent. Clicking on it reveals the issue.
I tracked it down to Window.open_window(): https://github.com/emoon/rust_minifb/blob/e5b5cc50d581494c38612c7dd835609d18e08eff/src/os/windows/mod.rs#L510-L514
Here different window flags are used to determine the size compared to when creating the window (line 552). According to https://stackoverflow.com/a/25879124 the same flags should be used for both WinAPI calls.

emoon:
Fixed by https://github.com/emoon/rust_minifb/pull/246

PR #246, opened 2021-04-02
Fix size calculation of non-decorated windows (#245)

swismer:
Use the correctly calculated window flags for both AdjustWindowRect and CreateWindowExW


emoon:
Thanks!

Issue #237, opened 2021-02-19
Reduce dependencies


emoon:
Currently when building for Linux one get a dependency list like this

```
  Compiling libc v0.2.74
   Compiling cfg-if v0.1.10
   Compiling proc-macro2 v1.0.19
   Compiling unicode-xid v0.2.1
   Compiling memchr v2.3.3
   Compiling pkg-config v0.3.18
   Compiling bitflags v1.2.1
   Compiling version_check v0.9.2
   Compiling lazy_static v1.4.0
   Compiling glob v0.3.0
   Compiling log v0.4.13
   Compiling xml-rs v0.8.3
   Compiling unicode-width v0.1.8
   Compiling regex-syntax v0.6.22
   Compiling cfg-if v1.0.0
   Compiling termcolor v1.1.2
   Compiling humantime v2.1.0
   Compiling bindgen v0.56.0
   Compiling strsim v0.8.0
   Compiling nix v0.17.0
   Compiling getrandom v0.1.14
   Compiling ansi_term v0.11.0
   Compiling vec_map v0.8.2
   Compiling rustc-hash v1.1.0
   Compiling peeking_take_while v0.1.2
   Compiling void v1.0.2
   Compiling lazycell v1.3.0
   Compiling shlex v0.1.1
   Compiling smallvec v1.4.2
   Compiling once_cell v1.4.0
   Compiling ppv-lite86 v0.2.8
   Compiling maybe-uninit v2.0.0
   Compiling byteorder v1.3.4
   Compiling downcast-rs v1.2.0
   Compiling cc v1.0.58
   Compiling crc32fast v1.2.0
   Compiling adler32 v1.2.0
   Compiling remove_dir_all v0.5.3
   Compiling libloading v0.6.7
   Compiling textwrap v0.11.0
   Compiling thread_local v1.1.0
   Compiling miniz_oxide v0.3.7
   Compiling nom v5.1.2
   Compiling clang-sys v1.0.3
   Compiling wayland-sys v0.27.0
   Compiling x11-dl v2.18.5
   Compiling aho-corasick v0.7.15
   Compiling deflate v0.8.6
   Compiling quote v1.0.7
   Compiling minifb v0.19.2 (/home/emoon/code/projects/rust_minifb)
   Compiling wayland-scanner v0.27.0
   Compiling png v0.16.7
   Compiling atty v0.2.14
   Compiling which v3.1.1
   Compiling raw-window-handle v0.3.3
[...]

PR #261, opened 2021-11-13
Fix #200

### LINKED ISSUE #200: Request feature to read typed keys
The crate already supports determining if a particular key is down or not, but this is not sufficient to accept text input in a natural way. There are two specific goals that can't be fulfilled:

 - When a key is pressed and held, there will be many typing events in a particular OS-defined pattern. To demonstrate this, hold down any letter on your keyboard and notice that the time between the first and second characters that are typed is different from after the 3rd, 4th, etc. Also, note that the exact intervals are OS-defined, and are often configurable in system settings.
 - In some cases, someone may want to type a character that requires a complicated series of key presses, in an OS-defined way. For example, consider the Linux compose key. When it is set to right alt, one can press right alt, then a, then ' to type á or right alt then ? then ? again to type ¿, where ? is typed with the shift key. There are many thousands of such combinations. 

A proper event or mechanism to look at _typed_ keys, as opposed to pressed keys, is needed to capture this complicated and OS-specific behavior. To provide this, I propose a new method on the `minifb::Window` struct, `fn get_next_typed_char() -> Option<char>` that, when called repeatedly, will give all characters that have been typed in the order that they were typed, then `None` once all types characters have been returned. This method could then be polled frequently (e.g. every frame) to get new information, that is simply passed along from the underlying OS-specific API. Note, however, that this is just one way to implement this, and that another method may be more convenient given the platform-specific underlying APIs (which are unknown to me).

john01dav:
#200 is important to me so I started working on a fix for it (it appears to not happen in windows, so I just need to make the X11 behavior more consistent with the correct behavior on Windows), but I ran into a road block that I described [here](https://users.rust-lang.org/t/why-does-xcreateic-in-xlib-return-null-over-ffi-but-not-in-c/67322). Is there any chance that I could get some help here? Also, I know that the current code on this branch is a mess -- don't worry, I'll clean it up quite a bit once it's working.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ features = [
 
 [features]
 default = ["wayland", "x11"]
-x11 = ["x11-dl", "xkb"]
+x11 = ["x11-dl", "xkb", "libc"]
 wayland = ["wayland-client", "wayland-protocols", "wayland-cursor", "tempfile", "xkb", "xkbcommon-sys"]
 
 [target.'cfg(not(any(target_os = "macos", target_os = "redox", windows)))'.dependencies]
@@ -49,7 +49,8 @@ wayland-cursor = {version = "0.28", optional = true}
 tempfile = {version = "3.2", optional = true}
 xkb = {version = "0.2.1", optional = true}
 xkbcommon-sys = {version = "0.7.5", optional = true}
-x11-dl = {version = "2.18.3", optional = true}
+x11-dl = {version = "2.19.1", optional = true}
+libc = {version = "0.2.107", optional = true}
 
 [target.x86_64-unknown-redox.dependencies]
 orbclient = "0.3.20"
--- a/src/os/posix/x11.rs
+++ b/src/os/posix/x11.rs
@@ -393,7 +393,6 @@ impl Window {
                 &mut attributes,
             );
 
-            //let empty_string = CString::new("").unwrap();
             let empty_string = b"\0";
             (d.lib.XSetLocaleModifiers)(empty_string.as_ptr() as *const i8);
 
--- a/src/os/posix/x11.rs
+++ b/src/os/posix/x11.rs
@@ -22,7 +22,7 @@ use crate::buffer_helper;
 use crate::mouse_handler;
 
 use super::common::Menu;
-use x11_dl::xlib::{XNInputStyle, XIMPreeditNothing, XIMStatusNothing, XNClientWindow, XNFocusWindow, XrmDatabase, KeyPressMask, KeyRelease, KeyReleaseMask, XKeyEvent, Status, XIC, XIM, Display, XErrorEvent, XEvent, ControlMask, KeySym};
+use x11_dl::xlib::{XNInputStyle, XIMPreeditNothing, XIMStatusNothing, XNClientWindow, XNFocusWindow, XrmDatabase, KeyPressMask, KeyReleaseMask, XKeyEvent, Status, XIC, XIM, XEvent, KeySym};
 
 // NOTE: the x11-dl crate does not define Button6 or Button7
 const Button6: c_uint = xlib::Button5 + 1;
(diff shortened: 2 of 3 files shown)


src/os/posix/x11.rs
```diff
@@ -384,6 +388,25 @@ impl Window {
                 &mut attributes,
             );
 
+            let empty_string = CString::new("").unwrap();
```
emoon, line 391: this can be replace with `let empty_string = b"\0";` instead

src/os/posix/x11.rs
```diff
@@ -1024,8 +1049,25 @@ impl Window {
...
-            callback.add_char(code_point);
+            callback.add_char(char);
+        }
+    }
+
+    unsafe fn code_point_to_char(&mut self, event: &mut XKeyEvent, code_point: u32) -> u32{
+        const BUFFER_SIZE: usize = 32;
+
+        unsafe {
+            let mut buff: Box<[u8]> = vec![0; BUFFER_SIZE].into_boxed_slice(); //TODO: extra heap allocation
```
emoon, line 1062: Also here remove this heap allocation with something like ` let mut buffer: [u8; 32] = [0; 32];`

john01dav: This was on my list of things to get done, I just wanted to get it working first. Normally I would not do a PR until I have such things cleaned up, but I need help, hence the early PR.

emoon:
Hey,

Thanks for the PR. What exactly do you need help with? I haven't done much in this area so I'm not sure of how much assistance I can be.

john01dav:
> Hey,
> 
> Thanks for the PR. What exactly do you need help with? I haven't done much in this area so I'm not sure of how much assistance I can be.

I need help figuring out why XCreateIC is returning null and fixing it.

emoon:
> I need help figuring out why XCreateIC is returning null and fixing it. (I can probably fix it myself once I know why it's happening)

I have never used this API so I'm not sure what could be wrong here

john01dav:
@emoon It's all ready to merge now. With regard to the bug I asked for help on, it turned out that I just made a small error in the C FFI for XCreateIC.

src/os/posix/x11.rs
```diff
@@ -384,6 +393,44 @@ impl Window {
                 &mut attributes,
             );
 
+            //let empty_string = CString::new("").unwrap();
```
emoon, line 396: Remove this line

Cargo.toml
```diff
@@ -49,7 +49,8 @@ wayland-cursor = {version = "0.28", optional = true}
 tempfile = {version = "3.2", optional = true}
 xkb = {version = "0.2.1", optional = true}
 xkbcommon-sys = {version = "0.7.5", optional = true}
-x11-dl = {version = "2.18.3", optional = true}
+x11-dl = {version = "2.19.1", optional = true}
+libc = {version = "0.2.107", optional = true}
```
emoon, line 53: What is `libc` used for? It would be good if no additional dependencies gets added.

john01dav: On line 120 of x11.rs I'm using it to set the current locale. This is needed for compose keys to work. I do not know why. To verify this, you can compile my branch on Linux (I'm testing on Debian 11), enable a compose key (a web search will show how to do this in whatever desktop environment you use), and then try to type a character with the compose key into the char_callback in my branch (I added a printout of what unicode character the code point corresponds to). If your compose key is right alt then right alt then a then ' types á. With the locale-setting line present, this works as intended. When the locale-setting line is commented out, no character is typed.

I don't think that it's ideal to have the locale set at a global basis like this, however I am not aware of any better solution. It's also worth noting that libc is already pulled in as a recursive/transitive dependency with the current dependency list, so adding it as a direc dependency is almost a no-op.

src/os/posix/x11.rs
john01dav: Done, sorry for missing that.

Cargo.toml
emoon: Alright. Fair enough, the reason I'm asking is that I want to try to reduce the big dependency list at some point as stated in #237

emoon:
Thanks!

PR #266, opened 2021-12-06
Bump versions

vemoo:


vemoo:
I have verified the examples work on linux under wayland and x11 and on windows, but I don't have a mac.

emoon:
Thanks! I can test on mac

vemoo:
I think the bump of `xkbcommon-sys` may have to be reverted, because I didn't realize that `xkb` also depends on it but a different version.

Also I found an [issue](https://github.com/meh/rust-xkbcommon-sys/issues/7) in the way the bindings are generated in `1.3`.

I will send a PR shortly.

PR #267, opened 2021-12-06
address clippy lints

vemoo:
Fix the clippy warnings that appear when running `cargo clippy` on linux. There may be other warnings on windows and/or macos.


emoon:
Great! Thanks

emoon:
Yeah, maybe I should have clippy running on the other platforms also

PR #269, opened 2021-12-12
macOS: Fix segmentation fault when resizing window

KaDiWa4:
Two obvious mistakes, but hard to find when you are only told that an invalid memory access happened in a specific place.


emoon:
Ouch! Great catch :)

PR #271, opened 2022-01-26
inputcallback: add a callback for key events

xobs:
Add a callback for key events. This includes all events such as Shift, not just text events. This allows for getting more complete information on keys as they're pressed.

An empty default handler is provided for backwards compatibility.


emoon:
Thanks!

PR #274, opened 2022-01-30
Add `set_icon()` method to set a window icon at runtime

xkevio:
I don't know if there is any interest in this functionality or if it was deliberately left out for simplicity but I went ahead
and tried my best at adding a `set_icon()` method under Windows and X11.

Wayland is not supported since they use `.desktop` files to link icons and executable files.
RedoxOS I am not sure they even have such a thing as icons yet.
MacOS I cannot test but it seems relatively easy from what I have seen.

Now for Windows (and probably MacOS) one can just use a file and a path to it, though not with X11.
X11 needs a `u64` ARGB buffer and I think implementing an "any image to ARGB" converter is out of scope, so
I tried to split up the functionality but probably didn't do it in the best way possible.

If there is any interest in adding this functionality, I hope someone can help adding the missing operating systems or just generally take a look at the code.


src/os/posix/x11.rs
```diff
@@ -606,6 +604,28 @@ impl Window {
         }
     }
 
+    #[inline]
+    pub fn set_icon(&mut self, icon: Icon) {
+        // XChangeProperty
+        let net_string = CString::new("_NET_WM_ICON").unwrap();
```
emoon, line 610: instead of doing this. Do `b"_NET_WM_ICON\0"` as that doesn't create a object on the heap.

xkevio: Good point, did that and fixed the errors in the workflow.
Also added an example in the `examples` folder since it might be weird to understand under Linux
but that can be removed of course.

emoon:
Thanks!

PR #276, opened 2022-02-02
Add getter for window position.

AndreasOM:
For a current project I wanted to store the layout of windows on the screen to reuse for the next session.

I added `get_position` to the window.
It is tested on macos, and windows - with a single display connected.
I created a stub for linux, which always returns 0,0. It compiles, but is untested due to lack of a system to test on.

It will need some rework for multi display systems, but I don't have one to do that.
I tried to keep the formatting, and style in line with the original.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -359,6 +359,24 @@ impl Window {
     }
 
     ///
+    /// Gets the position of the window. This is useful if you want
+    /// to store the position of the window across sessions
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
+    /// # use minifb::*;
+    /// # let mut window = Window::new("Test", 640, 400, WindowOptions::default()).unwrap();
+    /// // Retrieves the current window position
+    /// let (x,y) = window.get_position();
+    /// ```
+    ///
+    #[inline]
+    pub fn get_position(&self) -> (isize, isize) {
+        self.0.get_position()
+    }
+
+    ///
     /// Makes the window the topmost window and makes it stay always on top. This is useful if you
     /// want the window to float above all over windows
     ///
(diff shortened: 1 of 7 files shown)


xkevio:
This works for me under X11, only iffy thing is the `dummy_window` out variable that `XTranslateCoordinates` needs for potential error handling. I don't know how to just create an empty X11 `Window` struct.

```rust
#[inline]
pub fn get_position(&self) -> (isize, isize) {
    let (x, y);
    let (mut nx, mut ny) = (0, 0);

    // Create dummy window for child_return value
    let mut dummy_window = u64::MAX;

    let mut attributes: MaybeUninit<XWindowAttributes> = MaybeUninit::uninit();

    unsafe {
        let root = (self.d.lib.XDefaultRootWindow)(self.d.display);

        (self.d.lib.XGetWindowAttributes)(self.d.display, self.handle, attributes.as_mut_ptr());
        x = attributes.assume_init().x;
        y = attributes.assume_init().y;

        (self.d.lib.XTranslateCoordinates)(self.d.display, self.handle, root, x, y, &mut nx, &mut ny, &mut dummy_window);
    }

    (nx as isize, ny as isize)
}
```

emoon:
Thanks both of you!

Yeah, as for Wayland I looked at glfw and they also don't support it https://github.com/glfw/glfw/blob/master/src/wl_window.c#L880-L887 so I think we just have to live with returning `(0, 0)` on those platforms.

@AndreasOM can you bring in the changes (above) from @xkevio into the same PR?

AndreasOM:
Will be a bit busy the next few days, but should find some time on the weekend.
Will be a blind add of the linux part though.

AndreasOM:
Ah, well, skipped my coffee break, and just did it.

No promises it works, but it compiles.

emoon:
Thanks! I will merge it and give it a try :)

emoon:
I tested this today and it works just fine under Linux/X11 :)

PR #281, opened 2022-03-15
Add fileapi feature gate of winapi

arnabanimesh:
When building some apps, it throws an error at  `winapi::um::fileapi::GetFullPathNameW` as "could not find `fileapi` in `um`". This PR fixes the error.


emoon:
Thanks!

PR #290, opened 2022-04-15
wayland: Fix key character callback

### LINKED ISSUE #288: Strange keyboard behavior on wayland
I'm seeing strange keyboard behavior on wayland. It looks like keycodes are misinterpreted. This is most apparent running the `char_callback` example:

- keys `q`…`p` (first row of the keyboard) are ignored
- `d` maps to codepoint 32 (` `)
- `f` maps to codepoint 33 (`!`)
- `g` maps to codepoint 34 (`"`)
- … and so on, in qwerty order
- the numeric keyboard also returns keycodes, but the wrong ones

It seems like the key to character mapping is wrong, but I don't know anything about xkb internals as to what this could mean.

Context:
- Commit f42f516339cff04181bf2d984b48c543cd6f908e
- Swaywm
- Debian bookworm
- Rust 1.60.0
- Laptop with qwerty keyboard

I tried forcing to use Xwayland by unsetting `WAYLAND_DISPLAY`, but this runs into a different issue.

vmedea:
- Make the character callback use the key converted through the keymap, instead of trying to perform a direct conversion.

- Store a persistent keymap_state, instead of creating a new one every time. This makes sure that modifiers such as shift/capslock/ctrl are taken into account.

Closes #288.

### DIFF
--- a/src/os/posix/wayland.rs
+++ b/src/os/posix/wayland.rs
@@ -26,6 +26,7 @@ use wayland_protocols::xdg_shell::client::xdg_surface::XdgSurface;
 use wayland_protocols::xdg_shell::client::xdg_toplevel::XdgToplevel;
 use wayland_protocols::xdg_shell::client::xdg_wm_base::XdgWmBase;
 use xkb::keymap::Keymap;
+use xkb::state::State;
 
 use std::cell::RefCell;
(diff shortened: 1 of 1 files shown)


emoon:
Thanks!

Issue #291, opened 2022-04-16
Fails to compile on Raspberry Pi


wtfuzz:
I get the following error on a Raspberry Pi 400 running 64 bit bullseye.

Rust version:
```
$ rustc --version
rustc 1.60.0 (7737e0b5c 2022-04-04)
``` 

```
error[E0308]: mismatched types
    --> /home/fuzz/.cargo/registry/src/github.com-1ecc6299db9ec823/minifb-0.22.0/src/os/posix/wayland.rs:1092:25
     |
1092 |                         addr as *const i8,
     |                         ^^^^^^^^^^^^^^^^^ expected `u8`, found `i8`
     |
     = note: expected raw pointer `*const u8`
                found raw pointer `*const i8`

error[E0308]: mismatched types
   --> /home/fuzz/.cargo/registry/src/github.com-1ecc6299db9ec823/minifb-0.22.0/src/os/posix/x11.rs:400:41
    |
400 |             (d.lib.XSetLocaleModifiers)(empty_string.as_ptr() as *const i8);
    |                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u8`, found `i8`
    |
    = note: expected raw pointer `*const u8`
               found raw pointer `*const i8`

error[E0308]: mismatched types
   --> /home/fuzz/.cargo/registry/src/github.com-1ecc6299db9ec823/minifb-0.22.0/src/os/posix/x11.rs:625:62
    |
625 |                     (self.d.lib.XInternAtom)(self.d.display, net_string_ptr, xlib::False),
    |                                                              ^^^^^^^^^^^^^^ expected `u8`, found `i8`
    |
    = note: expected raw pointer `*const u8`
               found raw pointer `*const i8`

error[E0308]: mismatched types
   --> /home/fuzz/.cargo/registry/src/github.com-1ecc6299db9ec823/minifb-0.22.0/src/os/posix/x11.rs:626:62
    |
626 |                     (self.d.lib.XInternAtom)(self.d.display, cardinal_ptr, xlib::False),
    |                                                              ^^^^^^^^^^^^ expected `u8`, found `i8`
    |
    = note: expected raw pointer `*const u8`
               found raw pointer `*const i8`

For more information about this error, try `rustc --explain E0308`.
```

emoon:
Thanks for the report.

I wonder can you try to changes the places where this happens to do:

```
addr as _;
```
and 
```
cardinal_ptr as _
```

[...]

PR #292, opened 2022-04-16
Use coercion on types failing compilation on 64 bit raspbian bullseye

wtfuzz:
This addresses #291 where compilation failed on a Raspberry Pi with mismatched types errors.


emoon:
Thanks!

PR #295, opened 2022-05-06
[X11] Fix window names not supporting UTF-8

edgarogh:
Currently, UTF-8 characters don't render correctly in window titles (c.f. screenshots below). This is likely due to the fact that the X11 driver only uses `XStoreName`, which sets a name in a way that's not really intended for the user to see. This PR fixes that, making it possible to use UTF-8 characters (I tested diacritics and emojis) in the window title.

### How ?

In addition to the existing `XStoreName` call, I also call [`XChangeProperty`](https://linux.die.net/man/3/xchangeproperty) to set the [`_NET_WM_NAME`/`UTF8_STRING`](https://specifications.freedesktop.org/wm-spec/1.3/ar01s05.html) property. This adds one possible point of failure: `XChangeProperty` takes an `int nelements` as its last parameter, and the actual length of the string could be bigger than what a `c_int` can store. In the unlikely event that this happens, a specific error message is returned.

### Screenshots (context: I replaced a dash `-` with an em dash `—` in the fractal example)

Before:
fractal
After:
fractal2

---

Feel free to ask me to change a few things if needed, this PR is spontaneous, and I've never got to discuss the issue before.

### DIFF
--- a/src/os/posix/x11.rs
+++ b/src/os/posix/x11.rs
@@ -440,6 +440,22 @@ impl Window {
             }
 
             (d.lib.XStoreName)(d.display, handle, name.as_ptr());
+            if let Ok(name_len) = c_int::try_from(name.to_bytes().len()) {
+                let net_wm_name = d.intern_atom("_NET_WM_NAME", false);
+                let utf8_string = d.intern_atom("UTF8_STRING", false);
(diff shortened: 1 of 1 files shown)


emoon:
Great! Thanks :)

PR #299, opened 2022-07-26
Fix get_window_position for multiscreen setups on macOS.

### LINKED ISSUE #287: macOS: mfb_get_position uses the MainScreen size for calculations
If the window is on a secondary screen with different dimensions mfb_get_position and thus get_position will return wrong coordinates. 

Not sure what the best approach would be to get the screen the window is actually on.

And yes, I broke it. I just don't have a system with two screens.

AndreasOM:
Only lightly tested.
Should fix #287


emoon:
Great! Thanks :)

PR #304, opened 2022-12-26
set byposition flag when removing menus on windows

piksel:
I Could not get menus to be removed on windows, even when running the example.
Looking at the code, it looks like there is a flag missing as per https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-removemenu#parameters.

image

This PR simply adds the "By Position" flag that indicates that the menus should be removed by position, not by ID (which is what the `for i in 0..self.menus.len()` seem to indicate is the case).


emoon:
Thanks for the fix!

PR #308, opened 2023-01-28
Fix compilation of minifb's X11 binding for 32-bit

HBehrens:
FFI signature of `x11::xlib::XTranslateCoordinates` expects a `*mut c_ulong`.
Changing the cast to that  instead of `*mut u64` makes this compile on 32-bit
systems.

I can also confirm that the various examples run fine inside an arm32v7
docker image via X11.


emoon:
Thanks!

PR #321, opened 2023-06-13
Add `set_mouse_pos` function

arvsrn:
It only works on windows though

### DIFF
--- a/src/os/windows/mod.rs
+++ b/src/os/windows/mod.rs
@@ -909,6 +909,17 @@ impl Window {
         }
     }
 
+    pub fn set_mouse_pos(&mut self, top: u32, left: u32) {
+        if left < self.width as u32 && top < self.height as u32 {
+            unsafe {
+                let pos = self.get_position();
+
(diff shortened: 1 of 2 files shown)


src/os/windows/mod.rs
```diff
@@ -909,6 +909,17 @@ impl Window {
         }
     }
 
+    pub fn set_mouse_pos(&mut self, top: u32, left: u32) {
+        if left < self.width as u32 && top < self.height as u32 {
+            unsafe {
+                let pos = self.get_position();
+
+                winuser::SetCursorPos(left as i32 + pos.1 as i32, top as i32 + pos.0 as i32 + 30 /* height of title bar */);
```
emoon, line 917: I don't think hard coding height to 30 here is a good idea. The window may not have a title bar, or the user may have increased / decreased the font size of the title bar.

emoon:
Thanks for the PR. I can't merge this one yet tho as I strive to have the same functionally across all backends to not confuse users.

PR #329, opened 2023-10-24
raw-window-handle 0.5 (tested only on linux)

Noi42:
I made this for using minifb with vulkano.
I'm not sure it will work correctly on platforms other than Linux with x11.


emoon:
Thanks! I will test this some more after work and release a new version of the crate later today.

emoon:
I now noticed there is a 0.6 version of the crate. Any specific reason you didn't use that version?

Noi42:
Most crates still use 0.5 as far as I know. Same vulkano or ash that I use. But ash has a pull request on this topic [ash-window: Upgrade to raw-window-handle 0.6.0 #799](github.com/ash-rs/ash/pull/799) until recently, they were held up by a few other crates like winit.

PR #331, opened 2024-01-08
Moved runtime color bounds checks to compile time

StefanoIncardone:
made set_background_color function take u8 rgb values instead of usize and removed clamp function since it was only used to make sure that said usize color values were between 0 and 255, while u8 values are by definition guaranteed to be in that range


emoon:
Thanks!

PR #332, opened 2024-01-09
Improved code consistency and formatting

StefanoIncardone:
Added inline directives where it made sense and based on previous usage, namely implementation such as Window and Menu.
Moved freestanding functions inside impl blocks, thus deleting files with a single function definition, these were functions were the first argument could be changed with a "self". These functions were made sure to be marked as pub(crate) so not to leak them into the public interface of the crate.
Possibly found bug/typo in os/posix/wayland.rs is_key_released function.


emoon:
Thanks! I thought about doing a few of these changes before, but never got around to it :)

StefanoIncardone:
Happy to help where possible :)

PR #336, opened 2024-01-16
Fix disabling menu items on macOS

thisjaiden:
If a `NSMenu` does not have `setAutoenablesItems:NO` then any `NSMenuItem`s cannot be disabled. This fixes the functionality of disabling menu items on macOS.

Realistically down the line a test case should be added to the `menu` example for disabled menus, but it's not a huge concern and I'm not putting it in this PR.

I've tested this PR with the menu example which has no change (good!) and a personal project which is working as expected. This does not impact menu items that are enabled by default through minifb.


emoon:
Thanks!

Issue #339, opened 2024-01-26
Deprecate "limit_update_rate" in favor of "set_target_fps"


StefanoIncardone:
possibily replace [limit_update_rate](https://github.com/StefanoIncardone/rust_minifb/blob/11c7ac8b4b8138da241a7b36e5c8f219891ae63b/src/lib.rs#L527) with this implementation of "set_target_fps", since FPS is a more common and natural way to represent update intervals:

```rust
    /// Limits the FPS of polling for new events in order to reduce CPU usage.
    /// The problem of having a tight loop that does something like this
    ///
    /// ```no_run
    /// # use minifb::*;
    /// # let mut window = Window::new("Test", 640, 400, WindowOptions::default()).unwrap();
    /// loop {
    ///    window.update();
    /// }
    /// ```
    /// Is that lots of CPU time will be spent calling system functions to check for new events in a tight loop making the CPU time go up.
    /// Using `set_target_fps` minifb will check how many frames are left to reach the target FPS and there are any it will sleep for that amount of frames.
    /// This means that if more frames than the target happened (external code taking longer) minifb will not do any waiting at all so there is no loss in CPU performance with this feature.
    /// By default it's set to 250 FPS. Setting this value to None and no waiting will be done
    ///
    /// # Examples
    ///
    /// ```no_run
    /// # use minifb::*;
    /// # let mut window = Window::new("Test", 640, 400, WindowOptions::default()).unwrap();
    /// // Set the target rate to 60 fps, meaning events will be polled every ~16.6 ms
    /// window.set_target_fps(Some(60));
    /// ```
    #[inline]
    pub fn set_target_fps(&mut self, fps: Option<usize>) {
        match fps {
            Some(fps) => {
                let time = Some(Duration::from_secs_f32(1. / fps as f32));
                self.0.set_rate(time);
            }
            None => self.0.set_rate(None),
        }
    }
```

emoon:
I guess that would be fine, but I would prefer keeping the old one around for a while with a deprecation tag then.

StefanoIncardone:
> I guess that would be fine, but I would prefer keeping the old one around for a while with a deprecation tag then.

That was exactly what i was thinking, so i'm up for any suggestions on how you'd like to handle deprecations

emoon:
I guess having something like

```
#[deprecated(since = "0.26", note = "use set_fps_target instead, this function will be removed in the future."]
```
on the old function. Also given no time when it will be removed gives some flexibility

[...]

PR #340, opened 2024-01-26
Deprecate limit_update_rate

StefanoIncardone:
closing #339


emoon:
It would be good to change all the examples as well

StefanoIncardone:
> It would be good to change all the examples as well

Yes I plan on improving all of the examples and improving the relative documentation and comments in the future, since i found some examples to crash

emoon:
Alright. That sounds good! :+1:

PR #342, opened 2024-01-29
Improved examples

StefanoIncardone:
Improved consistency of examples and moved examples related resources out of the root folder and into the examples folder.
I also wanted to ask what to do with examples that crash (such as with [icon.rs](https://github.com/emoon/rust_minifb/blob/master/examples/icon.rs) on wayland: should we remove them or keep them for the future?


emoon:
As `set_icon` is a public API we should make sure that it doesn't crash under wayland even if wayland doesn't support setting icons on Windows.

emoon:
Merging this even if the icons example crashes on wayland. Great work once again!

StefanoIncardone:
> Merging this even if the icons example crashes on wayland. Great work once again!

There were quite a number of examples that did not work or crashed, due to missing implementation of internal methods, so these examples will "fix themselves" once the required functionalities get implemented

emoon:
Yeah, the problem is that I didn't implement the Wayland backend myself so it has some issues with some missing functionality. It all should be updated to the latest crates, but it requires a bunch of rewrite when doing that and I just haven't got around to do it.

Issue #347, opened 2024-03-21
Attempt to negate with overflow on latest rust nightly when running example code.


gillett-hernandez:
```powershell
$ rustc --version
rustc 1.79.0-nightly (a7e4de13c 2024-03-19)
```

example code:
```rust
use minifb::{Key, Window, WindowOptions};

const WIDTH: usize = 640;
const HEIGHT: usize = 360;

fn main() {
    let mut buffer: Vec<u32> = vec![0; WIDTH * HEIGHT];

    let mut window = Window::new(
        "Test - ESC to exit",
        WIDTH,
        HEIGHT,
        WindowOptions::default(),
    )
    .unwrap_or_else(|e| {
        panic!("{}", e);
    });

    // Limit to max ~60 fps update rate
    window.limit_update_rate(Some(std::time::Duration::from_micros(16600)));

    while window.is_open() && !window.is_key_down(Key::Escape) {
        for i in buffer.iter_mut() {
            *i = 0; // write something more funny here!
        }

        // We unwrap here as we want this code to exit if it fails. Real applications may want to handle this in a different way
        window
            .update_with_buffer(&buffer, WIDTH, HEIGHT)
            .unwrap();
    }
}
```

output:
```powershell
$ cargo run
thread 'main' panicked at C:\...\rust_minifb\src\os\windows\mod.rs:333:47:
attempt to negate with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: process didn't exit successfully: `target\debug\minifb_bug.exe` (exit code: 0xc0000409, STATUS_STACK_BUFFER_OVERRUN)
```

I have a fix, which i'll submit a PR for, but i'm not exactly sure why the fix works.

PR #348, opened 2024-03-21
fix negate with overflow, unsure why #[repr(C)] fixes it

gillett-hernandez:


gillett-hernandez:
fixes https://github.com/emoon/rust_minifb/issues/347

emoon:
Thanks! I'm unsure why this fixes it. The struct does get passed to `WndProc` and such, but everything is Rust side code so it shouldn't matter tho.

PR #350, opened 2024-04-11
Make scalar code plain C

Speykious:
I discovered that minifb uses and compiles non-Rust code for a few use cases, one of which is to make the image scaling functions optimized on Linux in all profiles.
I'm not keen on this approach, as it means that minifb is reliant on a separate compiler to build the project. As an example, I am using minifb for one of my projects, and someone who wanted to run it got a compiler error when compiling it on Linux. They didn't have a C++ compiler installed:
image

Ideally, I'd like to not depend on a C/C++ compiler on Linux.

One option is writing the scalar functions in Rust into their own crate, make minifb depend on that crate and ensure the profile for that dependency is always optimized:
```toml
[profile.dev.package.minifb-scalar]
opt-level = 3
```
although that means there is a second crate to be published on crates.io.

Another option is to just not optimize it, and instruct users of minifb to compile it with optimizations on all profiles:
```toml
[profile.dev.package.minifb]
opt-level = 3
```
although that means slower minifb if we want its debug info in our program. (There's also `debug = true` I guess?)

What are your thoughts on this matter?

This discussion aside, I made this PR to remove the need for a C++ compiler to be installed.
In my experience, it is much more likely for a C compiler to be installed on someone's Linux machine. I hope it's a bit useful :)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -45,7 +45,7 @@ fn main() {
     } else if !env.contains("windows") && !env.contains("wasm32") {
         // build scalar on non-windows and non-mac
         cc::Build::new()
-            .file("src/native/posix/scalar.cpp")
+            .file("src/native/posix/scalar.c")
             .opt_level(3) // always build with opts for scaler so it's fast in debug also
             .compile("libscalar.a")
     }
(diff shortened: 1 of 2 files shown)


Speykious:
Shoot, I didn't realize I was out of date...

Speykious:
I see. Yeah that makes sense
Thanks for merging!

PR #353, opened 2024-06-12
Fix `limit_update_rate`'s deprecation notice

Chubercik:
Previous deprecation notice pointed to a function that doesn't exist, this PR fixes that oversight.


emoon:
Thanks!

PR #359, opened 2024-08-24
fix image example buffer size calculation

JanNeuenfeld:
The calculation of the buffer size for a loaded image in the example code currently allocates a `Vec<u32>` that is exactly 4 times longer than needed.
This is due to `png::decoder::Reader::output_buffer_size()` returning the size of the buffer in bytes.
These bytes are later packed to `u32`s, which reduces the needed length of the `u32` buffer to `png::decoder::Reader::output_buffer_size() / 4`

### DIFF
--- a/examples/image.rs
+++ b/examples/image.rs
@@ -9,7 +9,7 @@ fn main() {
     decoder.set_transformations(Transformations::ALPHA);
     let mut reader = decoder.read_info().unwrap();
 
-    let mut buffer = vec![0u32; reader.output_buffer_size()];
+    let mut buffer = vec![0u32; reader.output_buffer_size() / 4];
 
     // View of pixels as individual subpixels (avoids allocating a second pixel buffer).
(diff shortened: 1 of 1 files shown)


JanNeuenfeld:
Also i went through the clippy suggestions. Please pay extra attention to the changed slice size calculations as I'm extremely tired at the moment and not sure if i overlooked reasons to not use `std::mem::size_of_val` in certain places.

emoon:
Thanks for the PR! LGTM 👍

emoon:
I will do some testing before merging this

JanNeuenfeld:
Okay that's embarrassing, sorry about that.
As far as I can tell everything still works on commit 3a4b729, so I'm pretty sure I messed up on dbe8eb1 when replacing manual slice size calculations.

emoon:
Sorry, I kinda screwed up when merging this and then reverting it. I wonder if you can open up a new PR with the correct fixes?

PR #362, opened 2024-08-29
Fix wasm compilation error

augustkline:
This PR fixes incompatible dependencies being brought into the build when building for the web.


emoon:
Thanks!

PR #363, opened 2024-08-29
Wasm Improvements

augustkline:
This PR makes some changes to how `Window`s are instantiated on wasm targets

- The `title` parameter in `Window::new()` has been replaced with `container`, which takes an element the canvas should be appended to
   - Allows more control over where and how to embed a minifb canvas in web projects, and still allows the user to change the window title with `set_title`
   - Rustdoc new window example has been updated for wasm32 target
- `set_cursor_visibility` now properly sets the the html canvas cursor style
- `is_active` now checks if the canvas element is focused

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,33 @@ maintenance = { status = "actively-developed" }
 
 [dev-dependencies]
 png = "0.17"
+console_error_panic_hook = "0.1.7"
+web-sys = { version = "0.3.56", features = [
+  "Blob",
+  "CanvasRenderingContext2d",
+  "CssStyleDeclaration",
+  "Document",
+  "Element",
+  "Gamepad",
+  "GamepadButton",
+  "GamepadEvent",
+  "Headers",
+  "HtmlCanvasElement",
+  "HtmlImageElement",
+  "ImageData",
+  "Navigator",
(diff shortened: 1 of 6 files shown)


emoon:
Hey,

Thanks for the PR! I wonder if it's possible to include an example that works with these changes also? It would be nice to keep the API same between desktop and wasm, but if it doesn't make sense from a web perspective I'm fine with the change.

iiisak:
> To avoid a breaking API change, `title` could be used to select a dom element by id...

Maybe WASM minifb should replicate what the other backends do, by creating the canvas itself. This way the title argument can just be used to set the title of the browser tab

emoon:
> Also, I noticed that the web feature is only used in 2 places (rate.rs and key_handler.rs) and could probably be replaced with a target = "wasm32" attribute, would it make sense to replace those and remove the web feature?

Sounds good to me

emoon:
Great! Thanks.

I wonder how hard would it be to add a build step for this in https://github.com/emoon/rust_minifb/blob/master/.github/workflows/ci.yml ? Just to make sure it doesn't break when doing other changes.

emoon:
Thanks for the PR. I will look into getting the build step for the CI running in another change.

Issue #365, opened 2024-09-13
Incorrect construction of a `Window`'s Win32 `RawWindowHandle` causes segmentation fault/pointer deref panic


jardhu:
When requesting a `RawWindowHandle` for a win32 `Window` (such as passing a `Window` to functions written with the `raw_window_handle` API), the instantiation of the `RawWindowHandle` performs some incorrect logic which results in either a panic after dereferencing a byte-misaligned pointer, or a segmentation fault.
```
thread 'main' panicked at C:\Users\jard\.cargo\registry\src\index.crates.io-6f17d22bba15001f\minifb-0.27.0\src\os\windows\mod.rs:518:53:
misaligned pointer dereference: address must be a multiple of 0x8 but is 0x2d61dca
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library\std\src/panicking.rs:645:5
   1: core::panicking::panic_nounwind_fmt::runtime
   ...
```

```
error: process didn't exit successfully: `target\debug\project.exe` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)

Process finished with exit code -1073741819 (0xC0000005)
```

There were three bugs which I've encountered and fixed with a workaround:

1. [The `HasWindowHandle` implementation for win32 assumes that the underlying window handle is a pointer to a `HWND` and dereferences it](https://github.com/emoon/rust_minifb/blob/39d8eee9333de3d5803d937afa3b495aaa4abad1/src/os/windows/mod.rs#L518C1-L521C11), when the handle is itself the `HWND` and simply needs to be casted to an `isize`. Hence, retrieving the `HWND` should look something like:
```rust
let hwnd = NonZeroIsize::new(self.window.unwrap() as isize).expect("invalid hwnd");
```
2. [Similarly, the code assumes the `HINSTANCE` handle is a pointer to an `HINSTANCE` and dereferences it.](https://github.com/emoon/rust_minifb/blob/39d8eee9333de3d5803d937afa3b495aaa4abad1/src/os/windows/mod.rs#L524) The `HINSTANCE` handle from `GetWindowLongPtr` needs to be casted to an `isize`.
[...]

PR #367, opened 2024-10-05
Fix incorrect deferencing of Windows window handles

Wumpf:
* Fixes https://github.com/emoon/rust_minifb/issues/365

Windows handles are pointers by type, but dereferencing these doesn't have any meaning and naturally may just crash (it does ;-)) - if you're lucky this is caught by Rust debug checks complaining about alignment, if you're not it's just a segfault.

Tested this on https://github.com/Wumpf/minifb_wgpu_web_and_desktop (at https://github.com/Wumpf/minifb_wgpu_web_and_desktop/commit/e39be06686b9b8662d27fd4cb0a3787dfaee172d)


emoon:
Great! Thanks :)

PR #368, opened 2024-10-05
[wasm] Fix raw handles, delay `getContext` call to be WebGPU compatible

Wumpf:
Also, fixed various warnings & added a CI step to check the wasm target.

The reason [`getContext`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext) have to be delayed here is that this stateful: Once `getContext` is called with one type of context (`2d`, `webgl` or `webgpu`), any future call with a different type will fail.
-> Previously, it was not possible to use the canvas with a WebGL/WebGPU context since `minifb` would call `.get_context("2d")` during initialization.

Tested this on https://github.com/Wumpf/minifb_wgpu_web_and_desktop (at https://github.com/Wumpf/minifb_wgpu_web_and_desktop/commit/e39be06686b9b8662d27fd4cb0a3787dfaee172d)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,10 +28,12 @@ jobs:
         os: [ubuntu-latest, windows-latest, macOS-latest]
     runs-on: ${{ matrix.os }}
     steps:
-    - if: matrix.os == 'ubuntu-latest'
-      run: sudo apt install libwayland-cursor0 libxkbcommon-dev libwayland-dev
-    - uses: actions/checkout@v2
-    - name: Build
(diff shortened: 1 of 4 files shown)


Wumpf:
ah the ci config didn't age well, have to fix this up

Wumpf:
done!
Couldn't have asked for a better example why the ci check is needed than the `unwrap_or(return false)` that slipped onto `master` meanwhile ;-)

emoon:
Awesome! Thanks for the fixes :)

PR #371, opened 2024-12-14
Fix windows hInstance for raw window handle

Wumpf:
..and make windows' hwnd a non-option, removing lots of 'unwrap'.

I think I actually broke the `hinstance` in a previous PR! Turns out that `GetWindowLongPtrW(raw_hwnd, GWLP_HINSTANCE)` always returned null. So instead, I make now sure to use the same hinstance that was used when creating the window in the  rist place.

I noticed this when trying to use wgpu with a Vulkan backend on windows which requires a valid hinstance. I didn't notice so far because wgpu always fell back to OpenGL because it (correctly) couldn't create a surface for Vulkan. But now when disabling OpenGL that error of course bubbled up.

### DIFF
--- a/src/os/windows/mod.rs
+++ b/src/os/windows/mod.rs
@@ -27,8 +27,7 @@ use winapi::{
         fileapi::GetFullPathNameW,
         libloaderapi, wingdi,
         winuser::{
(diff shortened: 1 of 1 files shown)


emoon:
Nice! Yeah, this looks much clear. Thanks for the cleanup :)

PR #374, opened 2025-01-21
Build - Remove use of `cfg!` and instead inspect `TARGET`

### LINKED ISSUE #373: Build script does not work correctly when cross-compiling
[Lines 24-30](https://github.com/emoon/rust_minifb/blob/9862ca89032d8ab442651ca4bf98dc38bf0df0e2/build.rs#L24-L30) of the build script use `cfg!` to determine if building for a linux target (to determine if the wayland/x11 features are required). However, that checks the compilation host (as it's evaluated when compiling the build script).

The correct behavior is done in the block starting at line 35, checking the contents of the `TARGET` script-runtime environment variable.

I noticed this issue while trying to cross-compile for windows on a linux host.

thepowersgang:
This fixes cross-compilation, as beforehand it would require the Linux windowing system features (X11 or Wayland) even when building for windows (on a linux host).

Tested locally using both linux->linux compilation (both that it builds with default-features, and errors without) and linux->windows (doesn't) error)

Fixes #373

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -19,21 +19,18 @@ fn main() {
     // target_arch is not working? OS FAMILY and ARCH variables were empty too
     // I think the cross-compilation is broken. We could take these from the environment,
     // since the build script seems to have a different target_arch than the destination.
-    let target = env::var("TARGET").unwrap_or("".to_string());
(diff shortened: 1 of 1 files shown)


emoon:
Thanks!

Issue #377, opened 2025-04-09
Released keys detection is broken on Wayland


woojiq:
How release keys detection works in minifb: it saves previous states of keys and latest state; on `get_keys_released` it can say if a key is released by checking prev and current states. This is ok. But implementation wise it's broken and it looks like it has never worked (I tried different commits up to 2022 year and it never worked).

Chain of events:

1. `update`/`update_with_buffer` is called.
2. Iterate over all input events: https://github.com/emoon/rust_minifb/blob/056c5afef1388247ca976f0edabd171a4393f3ed/src/os/posix/wayland.rs#L779
3. Process key events: https://github.com/emoon/rust_minifb/blob/056c5afef1388247ca976f0edabd171a4393f3ed/src/os/posix/wayland.rs#L794-L803
4. Set new key state: https://github.com/emoon/rust_minifb/blob/056c5afef1388247ca976f0edabd171a4393f3ed/src/os/posix/wayland.rs#L1095 https://github.com/emoon/rust_minifb/blob/056c5afef1388247ca976f0edabd171a4393f3ed/src/key_handler.rs#L37-L38

So far all good because `keys_prev` indeed contains previous states and `keys` contains latest state.

But what happens next is strange:

5. After processing all input events: https://github.com/emoon/rust_minifb/blob/056c5afef1388247ca976f0edabd171a4393f3ed/src/os/posix/wayland.rs#L955
6. And we overwrite `keys_prev` with latest data: https://github.com/emoon/rust_minifb/blob/056c5afef1388247ca976f0edabd171a4393f3ed/src/key_handler.rs#L73

As the resut we exit from `update`/`update_with_buffer` and now `keys_prev` and `keys` are equal, so `get_keys_released` is basically noop.

After I moved `self.key_handler.update()` call before processing new input events (`keys` -> `keys_prev` before processing new input events) it started to work. But there is few more lines of code in `KeyHandler:update`, so I'm afraid to break something else)

If you agree with me, I can open PR.

emoon:
I think this change looks good. If you want to open a PR for this it would be great. There are quite a bunch of issue with wayland https://github.com/emoon/rust_minifb/issues?q=state%3Aopen%20label%3A%22wayland%22 And it would be great to upgrade to the latest crates, but I just haven't any time to do it.

woojiq:
Opened PR. I'd like to upgrade crates and play with the Wayland protocol (no experience so far). But I don't have time right now either, xd 😄 
[...]

PR #378, opened 2025-04-11
Released keys detection is broken on Wayland

woojiq:
See commit message for the fix made and issue for context (don't want to duplicate information).

Fixes: #377

P.S. suggest merge commit as is since it contains valuable information.


emoon:
Thanks!

PR #390, opened 2025-11-05
implement get fps

Luca00casati:
there is doc to add, i will added tomorrow


Luca00casati:
so something like this?
    double GetTime(void);                                       // Get elapsed time in seconds since InitWindow() stolen from raylib

PR #391, opened 2025-11-25
: Replace unmaintained instant with web-time.

AndreasOM:
The `instant` crate is unmaintained.
This PR replaces it with `web-time`.

It works for all my projects, but **I do not have a wasm based project**. The tests compile for wasm, but I didn't run them!

web-time re-exports Duration and Instant from std::time on non wasm platforms, so it can handle the platform abstraction internally.


emoon:
Thanks!

PR #393, opened 2025-12-05
mac: improve numPad codes in KEY_MAPPINGS

gorsat:
Added several num pad keys that were missing in KEY_MAPPINGS for mac.


gorsat:
**NOTE**: I should have mentioned in the original PR that **this changes the mapping of keyCode 0x4c from ```Key::Enter``` to ```Key::NumPadEnter```**. This is consistent with the Windows implementation but if someone is relying on the existing mapping then this could be a breaking change for them. The other changes simply replace ```Key::Unknown``` with ```Key::NumPad...``` which seems unlikely to break anything but possibly could.

PR #399, opened 2026-03-28
Update minimum macOS version to resolve build issue

speelbarrow:
On macOS 26, encountering an issue where the library refuses to compile because some classes ('MTKView') are not available to target >10.11, but the build script in this repository was setting the minimum target to 10.10. Edited the build script, now the library builds fine.


emoon:
Thanks!

PR #400, opened 2026-05-21
Scale::FitScreen creates too big windows

StaterZ:
The current code has a window scaling bug for `Scale::FitScreen` that persists across almost all of the supported platforms.
The code makes an incorrect check for if the next scale increment would be too big, this pull request resolves this.

I have also standardized the algorithm across all currently supported platforms to ensure the issue is solved for others on other platforms.
The code also scaled in different increments across different platforms. This is now also standardized.


emoon:
Great! Thanks

PR #401, opened 2026-07-16
fix mac UAF from async buffer read in update_with_buffer_stride

ACK101101:
Supplement to demonstrate the bug and the fix! https://github.com/ACK101101/minifb-mac-update-uaf

### DIFF
--- a/src/native/macosx/MacMiniFB.m
+++ b/src/native/macosx/MacMiniFB.m
@@ -252,6 +252,7 @@ void* mfb_open(const char* name, int width, int height, uint32_t flags, int scal
     view.device = g_metal_device;
     view.delegate = viewController;
     view.autoresizingMask = NSViewWidthSizable | NSViewHeightSizable;
+    view.enableSetNeedsDisplay = YES;
     view.paused = YES;
     [window.contentView addSubview:view];
 
@@ -524,6 +525,7 @@ int mfb_update_with_buffer(void* window, void* buffer, uint32_t buf_width, uint3
     int state = generic_update(win);
     MTKView* view = (MTKView*)win->metal_view;
     [view draw];
+    win->draw_parameters->buffer = NULL;
 
     return state;
 }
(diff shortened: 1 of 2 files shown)


emoon:
Hey, Thanks for the PR!

~Can you describe more what this code fixes?~

The synchronous `[view draw]` does fix the UAF, since `replaceRegion:withBytes:` copies the buffer before `mfb_update_with_buffer` returns.

Two suggestions:

1. Clear the pointer after the draw, so any stray draw can never read freed memory (`drawInMTKView` already guards on a NULL buffer):
```objc
[view draw];
win->draw_parameters->buffer = NULL;
```

2. `view.paused = YES` with the default `enableSetNeedsDisplay = NO` means the view only draws on the explicit `[view draw]`. During a live resize drag AppKit's tracking loop blocks the app's update loop, so the window shows a stale stretched frame until the drag ends. With the pointer cleared as above, it's safe to also set `view.enableSetNeedsDisplay = YES` so resize/expose repaints still work.

ACK101101:
Makes sense, thank you for the suggestions! I am new to this. I ran into this bug doing my [first project in rust](https://github.com/ACK101101/surmise).
 
When trying out the change there, I see the window being drawn while drag moved, but not while being drag resized. I am unsure if I have it misconfigured or if there is something else I am missing.

emoon:
hum... I don't have access to a mac right now, but I can try it out to tomorrow.

emoon:
Sorry, I haven't managed to get around to check this yet. Will try as soon as I can

emoon:
I had a look and what I think we need to do is to remove this line
```C
view.enableSetNeedsDisplay = YES;
```
The issue is that to fix the resize issue with this enabled requires quite bit of more changes when this is enabled, which I think need it's own PR.

ACK101101:
Gotcha, just removed that line

emoon:
Cool. Does this solve the current resize issue?

ACK101101:
Sorry for the late response! I'm still seeing the resize issue on my project. Maybe that should be a separate issue still?

emoon:
Sounds good!

PR #428, opened 2026-08-18
fix: include payload in Error Display impl (#426)

### LINKED ISSUE #426: Display impl for Error discards the detail message
src/error.rs:15-24: `Display` prints only the variant name (\"Failed to Update\") and drops the payload string, so e.g. the detailed buffer-size explanation from `check_buffer_size` (src/lib.rs:1243) never reaches `println!(\"{}\", err)`, `err.to_string()`, or `Box<dyn Error>` reporting — it's only visible via `{:?}`. The payload should be included in `Display`.

chiliec:
## What

Closes #426.

The `Display` impl for `Error` printed only the variant name and discarded the `String` payload:

```rust
Error::MenuExists(_)   => write!(formatter, "Menu already exists"),
Error::WindowCreate(_) => write!(formatter, "Failed to create window"),
Error::UpdateFailed(_) => write!(formatter, "Failed to Update"),
```

So a detailed message — for example the buffer-size explanation built in `check_buffer_size` (`src/lib.rs`) — never reached `err.to_string()`, `println!("{}", err)`, or `Box<dyn Error>` reporting; it was only visible via `{:?}`.

## Fix

Bind and include the payload in `Display`:

```rust
Error::MenuExists(ref e)   => write!(formatter, "Menu already exists: {e}"),
Error::WindowCreate(ref e) => write!(formatter, "Failed to create window: {e}"),
Error::UpdateFailed(ref e) => write!(formatter, "Failed to Update: {e}"),
```

`Debug` is left unchanged (it already includes the payload).

## Tests

Added a unit test asserting `to_string()` includes the payload for each variant.

**Verified RED→GREEN:** with the old `Display`, the test fails (`left: "Failed to Update"`, `right: "Failed to Update: buffer too small"`); with the fix it passes.

## Validation (real results)

- `cargo test --lib display_includes_payload`: **passes**.
- `cargo fmt --check`: clean.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -16,9 +16,9 @@ impl fmt::Display for Error {
     fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Error::MenusNotSupported => write!(formatter, "Menus not supported"),
-            Error::MenuExists(_) => write!(formatter, "Menu already exists"),
-            Error::WindowCreate(_) => write!(formatter, "Failed to create window"),
-            Error::UpdateFailed(_) => write!(formatter, "Failed to Update"),
+            Error::MenuExists(ref e) => write!(formatter, "Menu already exists: {e}"),
+            Error::WindowCreate(ref e) => write!(formatter, "Failed to create window: {e}"),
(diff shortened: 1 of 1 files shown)


emoon:
Thanks!

PR #429, opened 2026-08-28
fix(wayland): shifted-keysym drops and a present-failure key desync

bearyjd:
Two real Wayland keyboard bugs found while building a remote-desktop client
(navette) that forwards physical key state over the network -- both make
the difference between "one dropped event" and "a stuck key" much sharper
than they would be for a purely local app, which is likely why they hadn't
surfaced before.

## 1. Shifted-level keysyms are silently dropped

`handle_key`'s match table only covers unshifted keysyms (lowercase
letters, unshifted punctuation). XKB resolves a key through whichever
shift level is currently active, so every shifted keysym -- capital
letters, `!`, `@`, `{`, etc. -- falls through to `_ => return` and never
reaches `set_key_state`.

Two consequences on a real keyboard:
- Shift+key does not register at all (the press itself is dropped).
- A press/release pair that straddles a Shift transition -- ordinary
  typing rollover: press a letter, press Shift for the next word, release
  the first letter while Shift is still down -- drops only the release.
  The key then reads as permanently held in `KeyHandler`'s state.

Fix maps each shifted keysym back to the same `Key` its unshifted form
already uses, matching the shift-invariant physical-key semantics every
other minifb backend has.

## 2. A failed present skips the key-repeat timer's advance

`update_with_buffer_stride` returned early via `?` on
`check_buffer_size`/`update_framebuffer` before ever calling
`self.update()` -- the only call site of `key_handler.update()`, which
advances each key's held-duration timer. A caller whose own error handling
for the `Err` doesn't also call `update()` (natural, since the point of
that branch is usually just log-and-retry) leaves the timer frozen for one
cycle. If a key had duration `0.0` (first poll after going down), the next
successful poll sees the same `0.0` again and reports it as freshly
pressed -- a duplicate keystroke with no visible cause tying it back to
the earlier failed present.

Fix runs `update()` unconditionally (captures the present result in a
closure first), so a failed present costs a frame but never desyncs input.

## Testing

Both fixes are exercised live in navette (a client that opens a real
window and forwards every keystroke over a real network connection to a
remote Wayland app), where the shifted-keysym fix was verified end-to-end:
capitals and shifted punctuation, previously untypeable, now work
correctly. `cargo test` and `cargo build --release` are clean on this
branch. I didn't add new unit tests here since `handle_key` takes a raw
XKB state pointer that isn't easily constructed in a unit test without a
real XKB context -- happy to add coverage if there's a preferred pattern
for this in the existing suite.

Two commits, one per fix, for easier review.

### DIFF
--- a/src/key_handler.rs
+++ b/src/key_handler.rs
@@ -49,7 +49,26 @@ impl KeyHandler {
         keys
     }
 
-    pub fn update(&mut self) {
+    /// Snapshots the current key levels as "previous", which is what
+    /// `is_key_index_released` compares against.
+    ///
+    /// A backend that applies platform key events itself must call this
+    /// *before* applying them: the release edge is `keys_prev && !keys`, so
+    /// snapshotting after a release has already been written to `keys` makes
+    /// both sides false and destroys the edge before the caller can read it.
+    pub fn snapshot_prev(&mut self) {
+        self.keys_prev.copy_from_slice(&self.keys);
+    }
+
+    /// Advances each key's held duration from the levels currently in `keys`.
+    ///
+    /// A backend that applies platform key events itself must call this
+    /// *after* applying them: `is_key_index_pressed` reports a key only while
+    /// its duration is exactly `0.0`, and a key pressed in this batch is still
+    /// at the initial `-1.0` until this runs. Advancing before the batch is
+    /// applied defers every press by a cycle, and loses it outright if the
+    /// release lands in the next batch.
+    pub fn advance_durations(&mut self) {
         self.delta_time = self.prev_time.elapsed();
         self.prev_time = Instant::now();
         let delta_time = self.delta_time.as_secs_f32();
@@ -64,8 +83,15 @@ impl KeyHandler {
             } else {
                 self.keys_down_duration[idx] = -1.0;
             }
-            self.keys_prev[idx] = self.keys[idx];
         }
+    }
+
+    /// Both phases together, for backends that apply their platform key
+    /// events outside the window this call spans (X11, macOS, Windows), where
+    /// the ordering the two phases exist to separate does not arise.
+    pub fn update(&mut self) {
+        self.snapshot_prev();
+        self.advance_durations();
     }
 
     #[inline]
--- a/src/os/posix/wayland.rs
+++ b/src/os/posix/wayland.rs
@@ -785,7 +785,10 @@ impl Window {
             self.should_close = true;
         }
 
-        self.key_handler.update();
+        // Snapshot before applying this batch, advance after it -- see the
+        // two methods' docs. This backend is the one that applies platform key
+        // events inline, so it is the one that has to split the phases.
+        self.key_handler.snapshot_prev();
 
         for event in self.input.iter_keyboard_events() {
             use wayland_client::protocol::wl_keyboard::Event;
@@ -838,6 +841,8 @@ impl Window {
                 _ => {}
             }
         }
+
+        self.key_handler.advance_durations();
 
         self.scroll_x = 0.;
         self.scroll_y = 0.;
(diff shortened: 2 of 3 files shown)


bearyjd:
Added a third commit to this branch: `0b54200` — `fix(wayland): split key_handler update so both key edges survive a slow poll`.

This one is related to the present-failure fix already in the PR but is a distinct bug, found while chasing a "typing repeats characters under load" report downstream.

`Window::update()` calls `key_handler.update()` *before* applying the keyboard events it just dispatched. `KeyHandler::update()` does two things that belong on opposite sides of that application:

- `keys_prev[i] = keys[i]` — what `is_key_index_released` compares against (`keys_prev && !keys`). Must be snapshotted **before** the batch, or a release written into `keys` makes both sides false and the edge is destroyed before the caller reads it.
- `keys_down_duration[i]` — `is_key_index_pressed` reports a key only while this is exactly `0.0`. Must be advanced **after** the batch, or a key pressed in it is still at the initial `-1.0` at read time.

Running both before the batch means a press is only reported if the key is *still down* at the next update. That's invisible at a fast poll (an ordinary keypress spans many cycles) and is why this hasn't bitten before. It breaks once a caller's poll cycle is slow enough that the release arrives in the next batch — the press is lost, an unpaired release is reported, and a key whose last observed edge was a press stays latched with `is_key_down()` still true.

Measured with a harness mirroring a real poll loop, identical input (20 taps, 50 ms hold), varying only cycle length:

| poll cycle | dup presses | unpaired releases | latched at exit |
|---|---|---|---|
| 8 ms | 0 | 0 | no |
| 60 ms | 0 | 10 | **yes** |
| 60 ms + this fix | 0 | 0 | no |

Worth noting the obvious one-line fix is wrong: moving `update()` wholesale to after the loop just trades lost presses for lost releases (measured 17 duplicate presses, 0 releases, still latched), because `keys_prev` then syncs to the post-release state. The phases have to be split.

Only the Wayland backend changes call sites; X11/macOS/Windows/Redox apply key events outside that window and keep calling `update()` unchanged. `cargo test` passes (35 tests); the 3 clippy warnings in `wayland.rs` are pre-existing (verified identical with and without the patch).

One residual left alone, inherent to the level-based key model: a press and release both landing inside one poll cycle net out and neither is reported — a dropped key, but no latch, so it can't become a stuck repeat.

emoon:
Thanks for the PR. The Copilot comment is valid, but that I will sorted that out in a follow up fix insetad.

bearyjd:
Thanks for the quick turnaround on this @emoon — merged inside a day, and it let us delete our fork immediately rather than carrying it. Much appreciated.

Since it might be useful back to you: we'd built a harness to characterise the third bug, and we re-ran it against master at `3a711ad` (i.e. including #430 and #431) before repinning, because both land in key handling. It injects 480 key edges through `virtual-keyboard-v1` into a nested sway, in four phases — single keys, rollover across Shift, sustained three-key overlap, deep four-key overlap — while the client polls at a deliberately slowed 60 ms cycle, which is what makes the edge-ordering bug reachable.

Result on current master:

| key | presses | releases |
|---|---|---|
| T | 84 | 84 |
| E | 50 | 50 |
| S | 40 | 40 |
| LeftShift | 35 | 35 |

Zero duplicate presses, zero unpaired releases, no key latched down at exit.

Also worth noting: #431 giving `Key` an explicit `#[repr(u8)]` formalises something downstream code was already depending on — we map minifb `Key`s to evdev codes, and that relied on the discriminants being what they now provably are. Good change to have made explicit.

One residual we documented on our side rather than pushing upstream, in case it's ever of interest: with a level-based key model, a press and release that both land inside a single poll cycle net out and neither edge is reported. That costs a dropped key, but importantly it can't latch one, so it degrades safely. We handled it by keeping our poll cycle short instead.

PR #432, opened 2026-08-29
Port Wayland backend to wayland-rs 0.31/0.32

### LINKED ISSUE #318: Upgrade wayland-client version
Minifb 0.24.0 (and current git) uses `wayland-client` at 0.29.5. Sadly, this produces a "future deprecation warning" because of old `wayland-client` dependencies.

    warning: the following packages contain code that will be rejected by a future version of Rust: nom v6.0.1
    note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`

Ideally, `minifb` would be updated to use `wayland-client` 0.30.1. I tried to update it myself, but there were too many breaking changes for me to work through it easily. For now, I'll just take the warning, which is against an example anyhow. But it would be great to fix it.

### LINKED ISSUE #408: Wayland: shm buffer release tracking is broken (tearing)
Two bugs in src/os/posix/wayland.rs defeat the wl_buffer release tracking:

1. Line 115: `get_buffer` selects a pool entry whose `buffer_state` flag is true, but nothing resets the flag to false when the buffer is re-attached. After a buffer's first Release the flag stays true forever, so from frame 3 on the same buffer is reused regardless of whether the compositor released it.
2. Lines 127-131: on resize, only `new_buffer.0` is stored; the new release-tracking `Rc` (`new_buffer.1`) is dropped, leaving `buffer_state` pointing at the destroyed old buffer's flag (stuck true).

Net effect: the client writes frames into shared memory the compositor may still be scanning out — tearing/partial frames, which is exactly what this machinery exists to prevent.

### LINKED ISSUE #409: Wayland: event channel can deadlock on event bursts
src/os/posix/wayland.rs:404-411 uses `mpsc::sync_channel(1024)`, with the senders invoked from `quick_assign` closures during `dispatch_pending` and the receiver drained via `try_iter` later in the same `update()` call — all on one thread.

`SyncSender::send` blocks when full. A 1000 Hz mouse plus a slow frame (or a long gap between `update()` calls) queues >1024 pointer Motion events; `dispatch_pending` then blocks in `send` with nothing draining — permanent hang.

Fix: use an unbounded channel, or drain into a Vec during dispatch.

### LINKED ISSUE #410: Wayland: keymap fd leak, xkb leaks, and panics on protocol errors
Robustness issues in src/os/posix/wayland.rs:

- Lines 1122-1146: the keymap fd from the compositor is never `close()`d after mmap (ownership transfers to the handler in wayland-client 0.29) — one fd leaked per Keymap event (layout change, seat hotplug).
- Lines 794-798: a second Keymap event overwrites `xkb_keymap`/`xkb_state` without unref (leak per layout change); `handle_keymap(...).unwrap()` panics on a malformed keymap and `KeymapFormat::NoKeymap` hits `unimplemented!` (line 1157) — compositor-supplied input shouldn't abort the process.
- Lines 767-771: `dispatch_pending(...).unwrap()` — a compositor disconnect during `update()` panics instead of setting `should_close`.
- Lines 95, 116, 139, 1211: buffer size math in `i32` (`size.0 * size.1 * 4`, and width*scale with `Scale::X32`) can overflow for large windows — protocol error or undersized pool.

emoon:
Moves the Wayland backend from the unmaintained 0.29 line to wayland-client 0.31 / wayland-protocols 0.32 / wayland-cursor 0.31.

The 0.29 event model (`Main<T>` + `quick_assign`) no longer exists; events now arrive through `Dispatch` impls on one state type, which queues keyboard and pointer events for `Window::update` to drain. The xkbcommon path is unchanged — level-0 keysym derivation, modifier tracking, held-key release bookkeeping and `add_char` all carried over, and `key_level_tests` still covers them.

`wayland-client` needs the `system` feature: `raw-window-handle` has to hand out a real `wl_surface*`/`wl_display*`, which only the libwayland backend can produce. `dlopen` still selects runtime loading vs linking, so the default build links neither libwayland nor libxkbcommon, as before.

Bugs fixed by the rewrite:

- Buffer release tracking. The flag was set on `wl_buffer::Release` and never cleared, and a resize dropped the new flag, so from frame 3 on a buffer was reused whether or not the compositor had released it. It is now the buffer's dispatch user data, cleared on attach. With every buffer held and the pool at its cap the frame is dropped rather than written into a buffer being scanned out.
- The shm file is sized before the pool is created and grown before the pool is — a mapping past EOF is a SIGBUS.
- `sync_channel(1024)` was filled and drained on the same thread, so an event burst could block dispatch with nothing draining it. Events queue in a `Vec` instead.
- The keymap fd is an `OwnedFd` and is closed; a keymap change unrefs the previous keymap and state.
- A malformed or non-XKB keymap is an error, not a panic. A compositor disconnect during `update()` closes the window instead of panicking.

Sizes and input coming from the compositor are validated rather than trusted:

- One `SurfaceSize` type carries a checked width, height and ARGB byte count, and `new()` is its only constructor, so holding one is the proof the size was checked and nothing downstream revalidates. Window creation, the configure path and the shm pool all take one. This replaces a first attempt at the check that could itself overflow: `i32::MAX * i32::MAX * 4` exceeds `i64::MAX`, so the guard meant to reject an oversized window panicked on it.
- Configure dimensions reached `scale_buffer`'s `i32` multiply unvalidated. They are now refused if unusable, and a configure may zero either axis on its own to leave that dimension to the client — a `(0, 600)` configure applies to the height instead of being dropped as a pair.
- An unknown `wl_pointer` `ButtonState` was read as a release and could clear a button still being held. It is ignored now, matching the key and axis handlers.
- A non-`WouldBlock` flush error was only logged, so a broken pipe left the window open despite the read and dispatch paths closing it.
- A keycode arrives as a compositor-supplied `u32` and was offset into xkb's numbering with a plain add, so a large one panicked in debug builds and wrapped to a low, valid-looking keycode in release ones.
- The keymap mapping was taken on trust. `mmap` does not check the length the compositor sends against the file behind the fd, so a short fd mapped cleanly and only faulted once xkbcommon read it, and a keymap with no terminator sent the parser's scan past the end of the mapping. Both are refused before parsing.
- A keymap the backend cannot use was logged and otherwise ignored, leaving the previous keymap and state live, so a compositor withdrawing its layout had subsequent keycodes translated through one it had already taken back. Translation stops instead, and keys held at that point are released.
- An `xdg_toplevel::Configure` size was published independently of the `xdg_surface::Configure` serial that closes the sequence, so a dispatch round ending between the two left a frame committing one configure's dimensions while acknowledging another's. The size is held pending until the serial arrives and the two are published together.

Drops nix 0.24, memoffset, wayland-commons and xml-rs; no duplicate crate versions remain.

Verified on Hyprland 0.56.2: examples run clean, keys and modifiers inject correctly via `wtype`, and `get_size` tracks the compositor's configures across a resize. 33 unit + 34 doctests pass. `cargo fmt --check` is clean and `cargo clippy --all-targets` reports exactly the same set of pre-existing warnings as master, across the wayland, wayland+x11, x11 and default feature sets.

Closes #318
Closes #408
Closes #409
Closes #410
Closes https://github.com/emoon/rust_minifb/pull/375

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -77,12 +77,15 @@ wayland = [
 ]
 
 [target.'cfg(not(any(target_os = "macos", target_os = "redox", windows, target_arch="wasm32")))'.dependencies]
-wayland-client = { version = "0.29", optional = true }
-wayland-protocols = { version = "0.29", features = [
+# `system` selects the libwayland-backed backend rather than wayland-rs'
+# pure-Rust one. It is required because `raw-window-handle` has to hand out a
+# real `wl_surface*`/`wl_display*`, which only that backend can produce.
+wayland-client = { version = "0.31", features = ["system"], optional = true }
+wayland-protocols = { version = "0.32", features = [
   "client",
-  "unstable_protocols",
+  "unstable",
 ], optional = true }
-wayland-cursor = { version = "0.29", optional = true }
+wayland-cursor = { version = "0.31", optional = true }
 tempfile = { version = "3.3", optional = true }
 x11-dl = { version = "2.19.1", optional = true }
 libc = { version = "0.2.107", optional = true }
--- a/src/os/posix/wayland.rs
+++ b/src/os/posix/wayland.rs
@@ -635,7 +635,7 @@ impl Window {
 
         let size = SurfaceSize::scaled(width, height, scale).ok_or_else(|| {
             Error::WindowCreate(format!(
-                "{}x{} at {}x scale is too large",
+                "{}x{} at {}x scale is not a usable window size",
                 width, height, scale
             ))
         })?;
--- a/src/os/posix/wayland.rs
+++ b/src/os/posix/wayland.rs
@@ -68,36 +68,65 @@ const KEY_MOUSE_BTN3: u32 = 274;
 const KEY_MOUSE_BTN8: u32 = 275;
 const KEY_MOUSE_BTN9: u32 = 276;
 
-/// Byte count of a `width` x `height` ARGB framebuffer, or `None` if the
-/// dimensions are not positive or the total does not fit the `i32` a
-/// `wl_shm_pool` size is sent as. Every path that accepts a size — window
-/// creation, a compositor configure, and the pool itself — goes through this,
-/// so `width * height` can never overflow downstream.
-fn framebuffer_bytes(width: i32, height: i32) -> Option<i32> {
-    if width <= 0 || height <= 0 {
-        return None;
-    }
-    i64::from(width)
-        .checked_mul(i64::from(height))
-        .and_then(|px| px.checked_mul(std::mem::size_of::<u32>() as i64))
-        .and_then(|bytes| i32::try_from(bytes).ok())
+/// A surface size with both axes positive and an ARGB byte count that fits the
+/// `i32` a `wl_shm_pool` size is sent as. `new` is the only constructor, so
+/// nothing downstream re-validates.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+struct SurfaceSize {
+    width: i32,
+    height: i32,
+    bytes: i32,
 }
 
-/// Which slot `BufferPool::get_buffer` should draw from.
+impl SurfaceSize {
+    fn new(width: i32, height: i32) -> Option<Self> {
+        if width <= 0 || height <= 0 {
+            return None;
+        }
+        let bytes = i64::from(width)
+            .checked_mul(i64::from(height))
+            .and_then(|px| px.checked_mul(std::mem::size_of::<u32>() as i64))
+            .and_then(|bytes| i32::try_from(bytes).ok())?;
+        Some(Self {
+            width,
+            height,
+            bytes,
+        })
+    }
+
+    fn scaled(width: usize, height: usize, scale: i32) -> Option<Self> {
+        let axis = |v: usize| i32::try_from(v).ok()?.checked_mul(scale);
+        Self::new(axis(width)?, axis(height)?)
+    }
+
+    /// A zero axis in an `xdg_toplevel::configure` means the compositor is
+    /// leaving that dimension to us, so the current value stands.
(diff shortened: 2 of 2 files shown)


emoon:
All three addressed in 1addb1e.

**Byte-count overflow** — correct, and the guard itself was the bug: `i32::MAX * i32::MAX * 4` is 1.8e19 against an `i64::MAX` of 9.2e18, so the check meant to reject an oversized surface panicked on it. Size validation now lives in one `framebuffer_bytes(w, h) -> Option<i32>` helper using checked multiplication, and every path that takes a size goes through it — window creation, compositor configure, and the pool — so `width * height` cannot overflow downstream. Covered by tests including the `i32::MAX` case.

**Flush errors** — right, a broken pipe on flush left the window open while the read and dispatch paths closed it. `try_dispatch_events` now returns `false` there too.

**Unvalidated configure** — right, the dimensions are compositor-controlled and went straight into `self.width`/`self.height`, so an oversized one reached `scale_buffer`'s `i32` multiply before `get_buffer` could reject it. A configure is now refused unless `framebuffer_bytes` accepts it; `0x0` is still handled as the normal "you choose" case.

emoon:
Both addressed in 7a6a825, along with consolidating the size checks that were accumulating.

**Zero configure axis** — correct, `xdg_toplevel::configure` zeroes each axis independently and the previous code rejected the pair, so `(0, 600)` was dropped instead of applying to the height alone.

**Unknown `ButtonState`** — correct, it evaluated to `false` and could clear a button still held. Now ignored, matching the key and axis handlers.

Rather than adding a fourth size check, validation is consolidated into a `SurfaceSize` carrying a checked width, height and byte count. `new()` is the only constructor, so holding one is the proof the size was checked; window creation, the configure path and the shm pool all take one and none of them revalidate. `width * height` no longer appears anywhere unchecked, and the zero-axis rule lives in a single `reconfigured()`.

33 unit + 34 doctests.

